PR #32, opened 2014-12-17
Burndown Data

jakobw:


app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
...
+		$this->tasks = $tasks;
+		$this->phabricator = $phabricator;
+		$this->days = $this->daysInSprint();
+	}
+
+	private function daysInSprint()
+	{
+		$days = [];
+
+		for ($day = strtotime($this->sprint->sprint_start);
```
JeroenDeDauw: Is there a reason you have public properties (as opposed to getters, and perhaps setters where applicable) in Sprint?

app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
...
+		     $day <= strtotime($this->sprint->sprint_end);
+		     $day += 60*60*24)
+		{
+			$days[] = $day;
+		}
+
+		return $days;
+	}
+
+	public function getDays($format = 'M j')
```
JeroenDeDauw: This is called both from within this class and from within a template. Slightly odd, not saying this is bad

app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
...
+	{
+		return array_map(function($day) use($format)
+		{
+			return date($format, $day);
+		}, $this->days);
+	}
+
+	public function closedPerDay()
```
JeroenDeDauw: It's hard for me to see what this method does quickly. The name can probably be better, and adding a doc that describes the format of what is returned would be helpful.

app/tests/unit/BurndownTest.php
```diff
@@ -0,0 +1,199 @@
...
+			$this->getMockBuilder('Phragile\TaskList')->disableOriginalConstructor()->getMock(),
+			$this->getMockBuilder('Phragile\PhabricatorAPI')->disableOriginalConstructor()->getMock());
+	}
+
+	public function dateCountProvider()
+	{
+		return [
+			['2014-01-01', '2014-12-31', 365],
+			['2014-12-01', '2014-12-14', 14],
+			['2014-11-28', '2014-12-2', 5],
```
JeroenDeDauw: No tests across a year boundary? Also, what about those with identical start and end date, or where the end date is before the start date?

app/tests/unit/BurndownTest.php
```diff
@@ -0,0 +1,199 @@
...
+						'oldValue' => 'resolved',
+						'newValue' => 'invalid',
+						'dateCreated' => '1418130000', // Dec 9
+					]
+				]
+			]
+		);
+
+		$closed = $burndown->closedPerDay();
+		$this->assertEquals(0, $closed['2014-12-09']);
```
JeroenDeDauw: You can use `assertSame` when comparing primitive types. Then this tell will fail if you get `null`, which it won't now

app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
...
+	{
+		$closedPerDay = array_fill_keys(array_merge($this->getDays('Y-m-d'), ['before', 'after']), 0);
```
JeroenDeDauw: I could not figure out why you are merging `['before', 'after']` together with `$this->getDays('Y-m-d')`. The resulting array is rather odd no? So I checked the test to see if it would tell me more, and while it reinforced my impression of primitive obsession, I'm still not entirely sure what this is about.

You can use associative arrays to describe objects, but this is not what arrays are for. Having different types of values in an array without keys is a flag, and having positional arguments in such arrays is an even bigger one.

app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
...
+		} elseif ($formattedDate < $this->sprint->sprint_start)
+		{
+			return 'before';
+		} else
+		{
+			return 'after';
+		}
+	}
+
+	private function closedTaskIDs()
```
JeroenDeDauw: These method names are more like property names. They don't start with a verb. `getClosedTaskIDs` is better.

app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
...
+	{
+		$taskTransactions = $this->phabricator->taskTransactions($this->closedTaskIDs());
+
+		return array_map(function($transactions)
+		{
+			return $this->findLastClosed($transactions);
+		}, $taskTransactions);
+	}
+
+	private function findLastClosed(array $transactions)
```
JeroenDeDauw: The first question to pop in my mind when seeing this method is "find closed _what_"? Tasks? Looking at the parameter list, and a quick look at the body, I can see it's transactions.

Next question: how do I call this? An array of transactions is needed, but what is a transaction. Looking at the body I can see that we again have an array thing instead of an actual type.

So now I wonder why there is no Transaction type. I'm also thinking you might want to have a Transactions collection, which would be a perfect home for a reduce method like this.

app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
+<?php
+namespace Phragile;
+
+class Burndown {
```
JeroenDeDauw: The name will probably be clearer if you make it a bit more noun like. Such as BurndownChart, if that is what this class actually is supposed to represent.

app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
...
+	// TODO: ideally these should come from a cached call to maniphest.querystatuses
+	private static $STATUS_OPEN = ['stalled', 'open'];
+
+	public function __construct(\Sprint $sprint, TaskList $tasks, PhabricatorAPI $phabricator)
+	{
+		$this->sprint = $sprint;
+		$this->tasks = $tasks;
+		$this->phabricator = $phabricator;
+		$this->days = $this->daysInSprint();
```
JeroenDeDauw: You did not define your fields

app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
...
+	}
```
JeroenDeDauw: You are doing work in the constructor by computing the `days` field. Not necessary bad, though definitely a code smell. Looks like the field is only accessed from one of the methods. Do you expect that method to be called multiple times? If not, then there is no need for this. If yes, then you can lazily compute the thing when needed.

app/Phragile/Burndown.php
```diff
@@ -0,0 +1,107 @@
...
+
+		for ($day = strtotime($this->sprint->sprint_start);
+		     $day <= strtotime($this->sprint->sprint_end);
+		     $day += 60*60*24)
+		{
+			$days[] = $day;
+		}
+
+		return $days;
+	}
```
JeroenDeDauw: Why is this method not in Sprint itself? It only uses fields of that class, so this suffers from the "Feature Envy" code smell

app/Phragile/TaskList.php
```diff
@@ -36,4 +38,38 @@ private function taskStatus(array $task)
...
+			} else
+			{
+				$acc[$task['status']] = [
+					'tasks' => 1,
+					'points' => $task['story_points']
+				];
+			}
+
+			return $acc;
+		}, ['total' => ['points' => 0, 'tasks' => '0']]);
```
JeroenDeDauw: why is the 0 for tasks a sting?

app/Phragile/TaskList.php
```diff
@@ -36,4 +38,38 @@ private function taskStatus(array $task)
 			? 'patch to review'
 			: $task['status'];
 	}
+
+	public function tasksPerStatus()
```
JeroenDeDauw: `getTasksPerStatus`

Also, return type doc

app/Phragile/TaskList.php
```diff
@@ -36,4 +38,38 @@ private function taskStatus(array $task)
...
+					'tasks' => 1,
+					'points' => $task['story_points']
+				];
+			}
+
+			return $acc;
+		}, ['total' => ['points' => 0, 'tasks' => '0']]);
+	}
+
+	public function findTaskByID($id)
```
JeroenDeDauw: parameter and return type docs

app/Phragile/Burndown.php
jakobw: They're not really different types of values. The resulting array describes the number of closed story points at a certain time. Tasks can be closed at a day of the sprint, `before` or `after`. I agree that it looks confusing and having all of it in one array wasn't as convenient as I thought anyway so I'll change that.

app/tests/unit/BurndownTest.php
jakobw: I added a test across the year boundary. The cases with invalid start and end dates are tested here: https://github.com/wmde/phragile/blob/master/app/tests/acceptance/create_sprint.feature#L29 and are caught by the Sprint validate method.

jakobw:
@JeroenDeDauw ping!

app/Phragile/PhabricatorAPI.php
```diff
@@ -77,4 +77,14 @@ public function createTask($projectPHID, array $task)
 			]
 		);
 	}
+
+	public function taskTransactions(array $ids)
```
JeroenDeDauw: missing `get`. I was wondering what `task` meant as a verb for a second here

app/models/Sprint.php
```diff
@@ -37,6 +38,39 @@ public function getPhabricatorError()
...
+	public function formatDays($format = 'M j')
+	{
+		return array_map(function($date) use($format)
+		{
+			return date($format, $date);
+		}, $this->getDays());
+	}
+
+	private function getDays()
+	{
```
JeroenDeDauw: You can make this simpler by doing

```
{
if ( $this->days === null ) $this->days = $this->computeDays();
return $this->days;
}
```

That way you have a method doing the caching and one doing the computation, and not one doing both.

bootstrap/start.php
```diff
@@ -59,7 +59,7 @@
 require $framework.'/Illuminate/Foundation/start.php';
 
 // Load libphutil for Phabricator Conduit calls
-require dirname(__DIR__) . '/vendor/libphutil/src/__phutil_library_init__.php';
+require_once dirname(__DIR__) . '/vendor/libphutil/src/__phutil_library_init__.php';
```
JeroenDeDauw: Wondering what the `dirname` is for

app/Phragile/BurndownChart.php
```diff
@@ -0,0 +1,109 @@
...
+		}, $taskTransactions);
+	}
+
+	private function findLastStatusChangeToClosed(array $transactions)
+	{
+		return array_reduce(
+			$transactions,
+			function($time, $transaction)
+			{
+				if ($transaction['transactionType'] === 'status'
```
JeroenDeDauw: You can extract this predicate to its own function that takes the $transaction as argument. Then the function name will make the code easier to read. Something like `statusChangedFromOpen`, if that indeed is what this if statement is about.

bootstrap/start.php
jakobw: Going up one level in the directory structure.

PR #47, opened 2015-01-08
Burndown frontend

jakobw:
@tobijat ping!


jakobw:
@snaterlicious Thanks for the review!
I added documentation for public methods and tried to address the structural points you mentioned. Let me know what you think!

snaterlicious:
Very nice improvements! There still is some data type documentation missing, but, as written, just suggestions.
You have created some scopes (which, basically, act as Singletons) that improve readabilty. Personally, I would like to use the constructors instead of `init()` functions. So, for example, in `pie_chart.js`, I would be able to call (maybe, it would even make sense, to move the DOM id parameter to the constructor):
```
statusPieChart = new StatusPieChart( scrapeStatusData() );
statusPieChart.render( '#pie', 250, 300 );
```
That concept would, for example, allow inheritance and improve modularity by another level. I guess, it would be possible to render multiple charts with different data (at least, that would be an indication for very good modularity). (Rendering multiple charts is possible at the moment as well, but there is no way to store references to the objects that are responsible for rendering the individual chart as `statusPieChart` acts as Singleton currently.) I am not sure whether all of that does, in fact, make sense as all of that simply may not be a requirement. So, again, just for consideration.
Structure of `StatusPieChart` would look kind of like the following:
```
var StatusPieChart = function( chartData ) {
  this._chartData = chartData;
};
StatusPieChart.prototype = {
  constructor: StatusPieChart,

  _width: null,

 [...]

  _setSVG: function( svgElementId ) {
    [...]
  },

  [...]

  render: function( closedPerDate, closedBeforeSprint, pointsInSprint ) {
    [...]
  }
};
```

PR #51, opened 2015-01-14
Sprint Snapshots

jakobw:
@JeroenDeDauw ping!


JeroenDeDauw:
Will try to review this tomorrow. Might not work, in case I'll do it on Friday.

app/models/SprintSnapshot.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+class SprintSnapshot extends Eloquent {
+	protected $fillable =  ['data', 'sprint_id'];
+
+	public function sprint()
```
JeroenDeDauw: `getSprint` or something like `getParentSprint`? Also, what does this return? A `Sprint` or some integer id?

app/models/Sprint.php
```diff
@@ -14,6 +15,11 @@ public function project()
 		return $this->belongsTo('Project');
 	}
 
+	public function sprintSnapshots()
```
JeroenDeDauw: `getSprintSnapshots`

This returns an array of ... ?

app/models/Sprint.php
```diff
@@ -92,4 +98,36 @@ private function createPhabricatorProject()
 
 		return true;
 	}
+
+	/**
+	 * @return bool Returns whether the snapshot was successfully created
+	 */
```
JeroenDeDauw: Do you need to do this return for some reason? It's a classical anti-pattern violating the [CQS principle](http://en.wikipedia.org/wiki/Command%E2%80%93query_separation)

app/Phragile/BurndownChart.php
```diff
@@ -9,13 +9,13 @@ class BurndownChart {
 	private $pointsClosedPerDay = null;
 	private $sprint = null;
 	private $tasks = null;
-	private $phabricator = null;
+	private $transactions = null;
 
-	public function __construct(\Sprint $sprint, TaskList $tasks, PhabricatorAPI $phabricator)
+	public function __construct(\Sprint $sprint, TaskList $tasks, array $transactions)
```
JeroenDeDauw: \o/

now It's not entirely obvious what type the parameter is now though. I'm assuming it's an array of `Transaction`, though it's good to document this.

app/Phragile/BurndownChart.php
```diff
@@ -38,26 +38,12 @@ private function calculatePointsClosedPerDay()
...
-				return $task['closed'];
-			})
-		);
-	}
-
 	private function getClosedTaskTimes()
 	{
-		$taskTransactions = $this->phabricator->getTaskTransactions($this->getClosedTaskIDs());
-
 		return array_map(function($transactions)
```
JeroenDeDauw: Should the parameter of the anon function not be called `$transaction`? Or is the transcations field an array of arrays?

app/models/Sprint.php
jakobw: It's a laravel thing which lets me access the SprintSnapshots of a Sprint through a dynamic property http://laravel.com/docs/4.2/eloquent#dynamic-properties

app/Phragile/BurndownChart.php
jakobw: It's an array of arrays containing the transactions for each task.

app/models/Sprint.php
JeroenDeDauw: So after quickly looking at this, I guess it returns a `Collection` which contains `SprintSnapshot` instances. Perhaps this is obvious if you know the framework. Still useful for people that don't.

PR #60, opened 2015-01-22
Do not use compact

JeroenDeDauw:


app/controllers/SprintsController.php
```diff
@@ -19,11 +19,17 @@ public function show(Sprint $sprint)
...
+		return View::make(
+			'sprint.view',
+			[
+				$sprint,
+				$sprint->project->currentSprint(),
+				$taskList,
+				new BurndownChart($sprint, $taskList, $phabricator->getTaskTransactions($taskList->getClosedTaskIDs()))
+			]
+		);
 	}
```
JeroenDeDauw: I think it's best to avoid `compact` due to it's high degree of magicness, and it being a rather unexpected way to go about passing along an array of local variables. Both these increase the chance of human error, and machine error (automated refactoring tools and static code analysers are likely to not handle this 100% correctly).

app/controllers/SprintsController.php
```diff
@@ -34,7 +40,7 @@ public function create(Project $project)
 			return Redirect::back();
 		}
 
-		return View::make('sprint.create', compact('project'));
+		return View::make('sprint.create', [$project]);
 	}
```
JeroenDeDauw: In this case using `compat` does certainly not save you any effort :)

app/controllers/SprintsController.php
jakobw: `compact('foo')` actually gives you `['foo' => $foo]`, so it does shorten things a little. I'm not 100% sure I want to change that here.

JeroenDeDauw: Oh. Does the using code need the key? Or can it also get `[$foo]`? If it needs the key, this change breaks the code.

jakobw: Yes, the keys of the array are the names of the variables in the respective view.

JeroenDeDauw: Well, my point about human error has been nicely illustrated then :)

PR #63, opened 2015-01-26
Create projects

jakobw:
@JeroenDeDauw ping!

### DIFF
--- a/.env.local.php.example
+++ b/.env.local.php.example
@@ -22,4 +22,6 @@ return [
     ],
     // The tag PHID that indicates whether a task is in review
     'REVIEW_TAG_PHID' => '',
+    // comma separated names of Phabricator user names that are allowed to create projects on Phragile
+    'PHRAGILE_ADMINS' => ''
 ];
--- a/app/models/User.php
+++ b/app/models/User.php
@@ -29,4 +29,9 @@ class User extends Eloquent implements UserInterface {
 
 		return true;
 	}
+
+	public function isAdmin()
+	{
+		return in_array(strtolower($this->username), array_map('trim', explode(',', $_ENV['PHRAGILE_ADMINS'])));
+	}
 }
--- a/app/models/User.php
+++ b/app/models/User.php
@@ -12,15 +12,25 @@ class User extends Eloquent implements UserInterface {
 
 	protected $fillable = ['phid', 'username'];
 
+	private $phabricatorURL;
+
 	public function setRememberToken($token) {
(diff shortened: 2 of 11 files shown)


.env.local.php.example
```diff
@@ -22,4 +22,6 @@ return [
     ],
     // The tag PHID that indicates whether a task is in review
     'REVIEW_TAG_PHID' => '',
+    // comma separated names of Phabricator user names that are allowed to create projects on Phragile
+    'PHRAGILE_ADMINS' => ''
```
JeroenDeDauw: Does Laravel not provide a more integrated way of doing this? And if so, why did you pick this approach? Just curious, not saying there is anything bad here.

app/models/User.php
```diff
@@ -29,4 +29,9 @@ public function certificateValid($certificate = null)
 
 		return true;
 	}
+
+	public function isAdmin()
+	{
+		return in_array(strtolower($this->username), array_map('trim', explode(',', $_ENV['PHRAGILE_ADMINS'])));
```
JeroenDeDauw: A bit too much complexity on one line IMO. I'd indent like

```php
return in_array(
    strtolower($this->username),
    array_map('trim', explode(',', $_ENV['PHRAGILE_ADMINS']))
);
```

Or whatever the equivalent with this CS is.

app/models/User.php
```diff
@@ -29,4 +29,9 @@ public function certificateValid($certificate = null)
 
 		return true;
 	}
+
+	public function isAdmin()
+	{
```
JeroenDeDauw: `$_ENV['PHRAGILE_ADMINS']` is global state usage. It'd be better to have this injected. Same goes for the `$_ENV['PHABRICATOR_URL']` above. Not sure to what extend the framework forces you into this, though it's definitely not good design.

.env.local.php.example
jakobw: To be honest I am not sure what the best practice is in this case. There is a class to handle configuration values using `Config::set` and `Config::get` which gets its information from the files in `app/config`. This seemed redundant to me since I would have to put the actual value in `.env.local.php` and then retrieve that value again using something like `Config.set('PHRAGILE_ADMINS', $_ENV['PHRAGILE_ADMINS'])` somewhere in `app/config`.
So I thought I might as well just put them here and retrieve them with `$_ENV`.

Using the Phabricator names there seemed the easiest way to do this for now without implementing actual user roles just for this one functionality. Arrays aren't supported for some reason so I went with a comma separated string.

JeroenDeDauw:
Looks good

PR #69, opened 2015-02-11
Show asignee

jakobw:
Ping @tobijat or @JeroenDeDauw ✨


resources/views/sprint/view.blade.php
```diff
@@ -143,6 +144,11 @@ class="hidden"
 						{{ $task['priority'] }}
 					</td>
 					<td class="points">{{ $task['story_points'] }}</td>
+
+					<?php $assignee = $assignees->getName($task['assignee']) ?>
```
JeroenDeDauw: $assigneeName

app/Phragile/AssigneeCollection.php
```diff
@@ -0,0 +1,45 @@
...
+class AssigneeCollection {
+
+	private $assignees;
+
+	public function __construct(PhabricatorAPI $phabricator, array $tasks)
+	{
+		$assigneePHIDs = $this->extractUniqueAssignees($tasks);
+		$this->assignees = $this->processAssignees(
+			$phabricator->getUserData(['phids' => $assigneePHIDs])
+		);
```
JeroenDeDauw: This prevents you from creating the collection if you already have the assignees.

app/Phragile/AssigneeCollection.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Phragile;
+
+class AssigneeCollection {
+
+	private $assignees;
```
JeroenDeDauw: Type hints for this field and for `$tasks` in the constructor are missing. Looks like they are both arrays of arrays, which is itself something to avoid. Still, indicating this with `array[]` is already more clear

app/Phragile/AssigneeCollection.php
```diff
@@ -0,0 +1,45 @@
...
+		return $assignees;
+	}
+
+	/**
+	 * @param string $phid
+	 * @return string
+	 */
+	public function getName($phid)
+	{
+		return isset($this->assignees[$phid]['userName']) ? $this->assignees[$phid]['userName'] : '-';
```
JeroenDeDauw: I personally prefer using array_key_exists when checking if an element is there or not, as I think this is less ambiguous about its intent. Just an FYI, I do not think this needs to be changed.

The `-` that is returned strikes me as a presentation concern. What if you want to replace the `-` in the page with `<i>none</i>`? Will you change it here?

JeroenDeDauw:
Have some more sparkles!

app/Phragile/AssigneeCollection.php
JeroenDeDauw: Actually, this is not used as a collection at all, so perhaps it does not matter it is not implemented as such. If your only public method is `getName($phid)`, then this to the outside looks more like a `AssigneeRepository`.

jakobw: That's a much better name, thanks!

jakobw:
@JeroenDeDauw yay for sparkles! I fixed all the things you mentioned :)

PR #88, opened 2015-05-04
Workboard support

jakobw:
@JeroenDeDauw Ping!
(Lots of code in here but I will try splitting things up more in the future)


JeroenDeDauw:
I already looked at most of this code, and will be doing so in the future anyway, so feel free to merge this. Better fix the test failure first though.

PR #94, opened 2015-05-11
Connect existing project

jakobw:
@JeroenDeDauw ping!


JeroenDeDauw:
> I a word.

lolwat? :)

app/Phragile/ActionHandler/SprintStoreActionHandler.php
```diff
@@ -0,0 +1,125 @@
...
+		if ($validation->fails())
+		{
+			Flash::error(implode(' ', $validation->messages()->all()));
+			$this->redirect = Redirect::back();
+		}
+	}
+
+	private function connectUserToPhabricator()
+	{
+		if ($this->previousActionFailed()) null;
```
JeroenDeDauw: Did you want to do a return here?

app/Phragile/ActionHandler/SprintStoreActionHandler.php
```diff
@@ -0,0 +1,125 @@
...
+	{
+		if (str_contains($errorMessage, 'Project name is already used'))
+		{
+			$this->connectWithPhabricatorProject();
+		} else $this->redirectBackWithError('Could not create a Phabricator project for this sprint.');
+	}
+
+	private function connectWithPhabricatorProject()
+	{
+		$this->sprint->connectWithPhabricatorProject($this->fetchPhabricatorProject());
```
JeroenDeDauw: I'm very dubious about using the no-curly-bracket notation for if and if-else statements. For the guard clauses above it's not that bad. However here I'd definitely use them and put the bodies of the thing on their own line.

app/Phragile/ActionHandler/SprintStoreActionHandler.php
```diff
@@ -0,0 +1,130 @@
...
+use Phragile\PhabricatorAPI;
+use \Flash;
+use Illuminate\Support\Facades\Redirect;
+
+class SprintStoreActionHandler {
+	private $userPhabricatorAPI;
+	private $botPhabricatorAPI;
+	private $redirect;
+	private $sprint;
+	private $user;
```
JeroenDeDauw: Be careful when using fields for internal state. Having fields set in the constructor are fine. Having fields set at the start of the only public method is also fine. Anything else requires caution with regard to state bugs. Following these rules, I can see `userPhabricatorAPI` and `botPhabricatorAPI` are fine. Since `sprint` and `user` are set in `performAction`, they are probably fine, since the two other public methods don't appear to interact with them. That leaves `redirect`. And indeed, it does look like there is potential for a state bug. If the service is used once, it will keep the value for `redirect`, and not reset it if it is used again.

JeroenDeDauw: I know that in this application no issue will occur at present, since this service is always invoked just once and then discarded. Still, I think it's good practice to avoid this kind of state in services wherever possible.

JeroenDeDauw:
Looks good to me, though please consider what I wrote for future code.

PR #95, opened 2015-05-13
Add functionality to connect Phabricator project by its Phabricator ID.

jakobw:
@JeroenDeDauw ping!
(not sure if you already gave green light for the previous PR so I didn't merge it yet.)

### DIFF
--- a/app/Phragile/ActionHandler/SprintStoreActionHandler.php
+++ b/app/Phragile/ActionHandler/SprintStoreActionHandler.php
@@ -3,6 +3,7 @@ namespace Phragile\ActionHandler;
 
 use Phragile\PhabricatorAPI;
 use \Flash;
+use \Sprint;
 use Illuminate\Support\Facades\Redirect;
 
 class SprintStoreActionHandler {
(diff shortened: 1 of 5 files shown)


app/Phragile/ActionHandler/SprintStoreActionHandler.php
```diff
@@ -3,6 +3,7 @@
 
 use Phragile\PhabricatorAPI;
 use \Flash;
+use \Sprint;
```
JeroenDeDauw: You do not need the preceding backslash

app/Phragile/ActionHandler/SprintStoreActionHandler.php
```diff
@@ -41,6 +46,34 @@ private function validate()
 		}
 	}
 
+	private function trySprintCreationFromPhabricatorID()
+	{
+		if ($this->previousActionFailed()
+			|| !is_numeric($this->sprint->title)
+			|| !is_int($this->sprint->title + 0)) return;
```
JeroenDeDauw: You sure the last part of this works correctly in all cases? I've used `ctype_digit` in the past. It's a bit odd that you do not know here if the thing is a string or an int.

PR #108, opened 2015-06-05
Clicking the same filter twice should show all items again

Benestar:
Fixes https://phabricator.wikimedia.org/T100193


jakobw:
Nice! I'll have a look at this today.

jakobw:
@Benestar Works perfectly!
Two small things:

1. Could you run `gulp --production` locally to generate the minified js file for this and then commit that too?
2. I think storing the filter status as a string will become tricky once we try to implement [multiple filters](https://phabricator.wikimedia.org/T100191). I haven't made up my mind about this yet and it works for now, so the current solution is fine.

Thanks a lot!

Benestar:
Well, `search` won't work with multiple filters. This should be done using `filter`. Are you already working on something like that? Otherwise I will just make a PR for that as well and change the code to use `filter`

Benestar:
Closing in favour of #110

PR #109, opened 2015-06-08
Snapshots command

jakobw:
ping @WMDE-Fisch or @KaiNissen 

This creates an [artisan command](http://laravel.com/docs/5.0/commands) which allows us to create snapshots for all Phragile sprints at once (e.g. daily via cron jobs) by executing `php artisan snapshots:create`.


WMDE-Fisch:
looks sane :+1:

PR #110, opened 2015-06-10
Support multiple filters

Benestar:
This fixes [T100191](https://phabricator.wikimedia.org/T100191) and [T100193](https://phabricator.wikimedia.org/T100193)

<s>Note that filtering the "status" column is broken because the `class="status"` is assigned to the wrong element (the `<td>` instead of the `<span>`).</s>


jakobw:
Looks good except for whitespace (spaces in the js files) and it would be awesome if you could  `gulp --production` and then commit the minified files that are generated.

Benestar:
Did you come to a decision how to handle whitespaces in javascript files? Should I make them spaces? Or do you want to make them tabs anyways?

jakobw:
For now it should be spaces in the js files.

Benestar:
Somehow, `gulp --production` doesn't like me and throws errors on my machine :S Can you please run that  before merging?

jakobw:
Done and merged \o/

PR #111, opened 2015-06-10
Make the 'total' link also clear the filters

Benestar:
Clicking that link no longer works with #110


jakobw:
Merged!
Thanks a lot! :+1:

PR #115, opened 2015-07-03
Add migration for allowing nulls in closed_statuses

wmde-manicki:
Reason for the change is that `closed_statuses` does not allow for null values currently but the default value in schema is actually null. The value can only be set once project has been created and contains a sprint. Default null value causes crash when trying to add new project in Phragile.

I am not happy with the down function (dropping the column). If there are better way to revert setting column to nullable, feel free to comment.
Drawbacks of the method here are that rollback will delete all current values of `closed_statuses` and (at least in MySQL) change the order of columns (placing `closed_statuses` as a last column).


wmde-manicki:
this is not solving all problems with undefined value of `closed_statuses` (see behat tests failing), so this should be done another way.

wmde-manicki:
The whole problem was caused by using `STRICT_TRANS_TABLES` mode of MySQL as has been suggested by Phabricator. Turning it off resolved the issue. This patch is then completely irrelevant.

PR #118, opened 2015-07-09
Add missing parameters and settings to Elixir functions

wmde-manicki:
New version of compiled Javascript and CSS files has not been included in `burnup_frontend` branch. This is actually fine (in my opinion) but in that case I had to be regenerate those compiled files on my own.

When I run gulp I get css stylesheet compiled but to the wrong file (`public/css/app.css` instead of `public/css/style.css`). Gulp has also generated `app.css.map` which, as I understand, is not used by Phragile. Javascript files fail to compile as Gulp could not found input Javascript files.

This problem may be resolved by adding some paths to the `Gulpfile`. CSS map file is also no longer generated (it could be also added to .gitignore file but it seems to me more reasonable not to create it at all).

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -1,7 +1,9 @@
 var elixir = require('laravel-elixir');
 
+elixir.config.sourcemaps = false;
+
 elixir(function(mix) {
-    mix.less('style.less');
+    mix.less('style.less', 'public/css/style.css');
 
     mix.scripts([
         '../assets/bower_components/d3/d3.js',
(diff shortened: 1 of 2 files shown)


jakobw:
Alright, this seems to have been a problem with different Gulp versions. Running `gulp` initially failed for me with these changes but it worked `npm update`. I guess this means we should require our Gulp version to be >=3.9 in the `package.json`?

wmde-manicki:
I would also suspect that it may has been Elixir that had changed (I believe this `npm updated` has also updated the Elixir version of yours, right?).
Elixir is apparently still requiring Gulp's version to be `^3.8.8`, so we may probably not need to require 3.9.
On the other hand, I failed to track when did actually possible changes to this particular part of Elixir happened, they don't document version releases, or I am too blind to find it.

wmde-manicki:
10 seconds later: It may be this one: https://github.com/laravel/elixir/commit/6f32b1b9ca185924a1bb5f4997e11c2e5e91ffb4, at least for the directory part of the fix

jakobw:
True, that seems to be what happened. I can't remember when I last ran `npm update` and Elixir is pretty new.
I just noticed that Laravel now wants Elixir to be ^2.0.0 so maybe that is what we should specify in the package.json instead. I'll change that quickly.

wmde-manicki:
And now I have also found out that there is no need to disable generating the CSS sourcemap files (`elixir.config.sourcemaps = false;`). They are not generated when running gulp in production mode, and generated otherwise. It seems reasonable.
I will update the patch in a minute

jakobw:
Looks good to me now

wmde-manicki:
Yeah, I got no further comments on that

PR #114, opened 2015-07-02
Burnup frontend

jakobw:

PR #121, opened 2015-07-13
Export graph data as JSON

wmde-manicki:
This will resolve https://phabricator.wikimedia.org/T100184 and depends on https://github.com/wmde/phragile/pull/114

Note that output format is not exactly the same as suggested in  https://phabricator.wikimedia.org/T100184 (field containing number of story points on non-open tasks at the given date is called`points` instead of `done`). The format could be changed accordingly once it is eventually specified.

@gbirke @Andrew-WMDE this is the way we were more or less going to take during our short session. Feel free to comment!
It uses the route suggested by @gbirke


wmde-manicki:
Moved to https://github.com/wmde/phragile/issues/123

PR #126, opened 2015-08-04
Fix project column bug.

jakobw:
The bug appeared whenever the last column move of a task happened on a worboard other than the sprint's own workboard and the column's name was the same as one of sprint workboard columns (e.g. Done).
Phragile would then use the PHID of the other workboard's column too look up at what date a task was moved on the sprint workboard and not find anything. The tasks' statuses were still correct since the names were the same but the `closedTime` lookup failed.


tobijat:
This was bug https://phabricator.wikimedia.org/T102777

PR #129, opened 2015-08-11
Sprint settings

jakobw:
Task on Phabricator: https://phabricator.wikimedia.org/T100187

### DIFF
--- a/app/Http/Controllers/SprintsController.php
+++ b/app/Http/Controllers/SprintsController.php
@@ -59,4 +59,12 @@ class SprintsController extends Controller {
 
 		return $actionHandler->getRedirect();
 	}
+
+	public function updateSettings(Sprint $sprint)
+	{
+		$sprint->update(Input::only('ignore_estimates'));
+
+		Flash::success('The sprint settings have been updated');
+		return Redirect::back();
+	}
 }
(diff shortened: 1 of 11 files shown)


wmde-manicki:
I have tested this on my local Phragile instance and seems to be working fine. Code changes are also sane. I would merge this but I am just not sure if the following thing was intended (sorry for the lengthy description):

I have created a new sprint for the project that already has some finished sprints. Added some tasks to the current sprint, some had story points estimates, some did not. Returned to Phragile, current sprint's graph has been plotted based on story point estimates (tasks with no estimates were counted as having zero SPs). Everything looked fine. Then turned the "ignore estimates" setting on. Current sprint's graph is now based on number of task. So the change is working, yeah! Then I switched to some finished sprints of the same project. Graph has then been plotted using snapshot so it is based on story points estimates - that's what I'd expect. But in the task list below the graph task each task is having one in the story points column as if the "ignore story point estimates" was turned on for this sprint as well. I then went to "Sprint settings" of this finished sprint and the ignore setting was in fact turned on although I have not switch for this particular sprint. At this point I started wondering if this is intended and am I understanding the scope of sprint settings wrong?

jakobw:
@wmde-manicki The settings being global issue had something to do with using sprint id vs the sprint's phabricator id. I'm surprised it worked at all. Well spotted!
I also changed the behavior of snapshot so that they also consider the `ignore_estimates` setting. It makes the charts look weird when you switch back and forth but that shouldn't happen and I think it's more consistent now.

wmde-manicki:
Looks good now!

PR #135, opened 2015-08-13
Add ability to delete a sprint

Benestar:
Fixes https://phabricator.wikimedia.org/T108295

In this version it is only checked that the user is logged in. Further checks if the user is allowed to delete the sprint should be added in another pull request.

Strangely, the acceptance tests all fail on my machine so perhaps someone should run my new test before merging this :-)

### DIFF
--- a/tests/acceptance/bootstrap/FeatureContext.php
+++ b/tests/acceptance/bootstrap/FeatureContext.php
@@ -459,4 +459,12 @@ class FeatureContext extends MinkContext implements Context, SnippetAcceptingCon
 	{
 		PHPUnit::assertCount(intval($number), $this->responseJSON()->$key);
 	}
+
+	/**
+	 * @Then the sprint :sprint should not exist
+	 */
+	public function theSprintShouldNotExist($sprint)
+	{
+		PHPUnit::assertNull(Sprint::where('title', $sprint)->first());
+	}
 }
--- a/tests/acceptance/sprint_overview.feature
+++ b/tests/acceptance/sprint_overview.feature
@@ -42,3 +42,10 @@ Feature: Sprint Overview
     And I should see "2014-12-06"
     And I should see "2014-12-07"
     And I should see "2014-12-08"
+
+  Scenario: Delete sprint
+    Given a sprint "Sprint 42" exists for the "Wikidata" project
+    And I am logged in
+    When I go to the "Sprint 42" sprint overview
+    And I click "Delete sprint"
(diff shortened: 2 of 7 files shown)


tobijat:
We could do the restriction to admins in a separate PR but they should be merged together with this one. Otherwise Phragile would be in a state where every logged in user could delete sprints since master gets deployed continuously .

wmde-manicki:
Looks fine for me. Also, I have had the acceptance test for deleting the sprint run and passed on my local instance.
Once sprint is removed any snapshots related to the sprint are **not** removed. I believe it is what is intended.
All in all, I would merge it but having @tobijat's comment in mind would wait for the other patch restricting the right to delete sprints.

tests/acceptance/sprint_overview.feature
```diff
@@ -42,3 +42,10 @@ Feature: Sprint Overview
     And I should see "2014-12-06"
     And I should see "2014-12-07"
     And I should see "2014-12-08"
+
+  Scenario: Delete sprint
+    Given a sprint "Sprint 42" exists for the "Wikidata" project
+    And I am logged in
+    When I go to the "Sprint 42" live page
+    And I click "Delete sprint"
+    Then the sprint "Sprint 42" does not exist
```
Benestar: I just wondered if this should be `Then the sprint "Sprint 42" **should** not exist` because `the sprint xxx does not exist` is defined as a `Given` clause. Does anyone know?

jakobw: It should definitely not use the `theSprintDoesNotExist` as it actively deletes the sprint. A new Then-step would be the ideal way to go.

Benestar: Ok, I will try to create one

tests/acceptance/bootstrap/FeatureContext.php
```diff
@@ -459,4 +459,12 @@ public function thePropertyShouldContainElements($key, $number)
 	{
 		PHPUnit::assertCount(intval($number), $this->responseJSON()->$key);
 	}
+
+	/**
+	 * @Then the sprint :sprint should not exist
+	 */
+	public function theSprintShouldNotExist($sprint)
+	{
+		PHPUnit::assertFalse(Sprint::where('title', $sprint)->first());
```
jakobw: Should be `assertNull`

Benestar: Fixed

PR #138, opened 2015-08-14
Adjust acceptance tests after having the way of generating sprint snapshots

wmde-manicki:
Since https://github.com/wmde/phragile/commit/97c6f52560bdcbe30b9c424a2ad035a3986cb87c Artisan's `snapshots:create` command is generating snapshots for active sprints only.
This should be also considered in the acceptance test.

### DIFF
--- a/tests/acceptance/bootstrap/FeatureContext.php
+++ b/tests/acceptance/bootstrap/FeatureContext.php
@@ -412,11 +412,15 @@ class FeatureContext extends MinkContext implements Context, SnippetAcceptingCon
 	 */
 	public function iShouldHaveCreatedOneSnapshotForEachSprint()
 	{
-		$numberOfSprints = Sprint::count();
+		$numberOfActiveSprints = count(array_filter(Sprint::all()->all(), function($sprint)
+			{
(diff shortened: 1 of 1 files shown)


tests/acceptance/bootstrap/FeatureContext.php
```diff
@@ -412,11 +412,15 @@ public function iExecuteArtisan($command)
 	 */
 	public function iShouldHaveCreatedOneSnapshotForEachSprint()
 	{
-		$numberOfSprints = Sprint::count();
+		$numberOfActiveSprints = count( array_filter( Sprint::all()->all(), function($sprint)
```
jakobw: No spaces after `count(` and `array_filter(` and after the closing parenthesis.

jakobw:
Looks good except for the minor code style thing!

tests/acceptance/bootstrap/FeatureContext.php
wmde-manicki: Yeah, I've just subconsciously stuck to the style used in other projects. Fixed this

PR #143, opened 2015-08-18
Don't override title if already set

Benestar:
Fixes https://phabricator.wikimedia.org/T105899


jakobw:
Looks good but needs rebase

PR #144, opened 2015-08-18
Fix sprint exists step in behat tests.

jakobw:
Before this fix sprints were missing the `phid` and `phabricator_id` whenever they got deleted and re-added by the behat tests. This fixes the issue by fetching the missing attributes from Phabricator.


wmde-manicki:
All acceptance tests passing now!

PR #141, opened 2015-08-18
Sprint not found

jakobw:
This solves https://phabricator.wikimedia.org/T104455

When a sprint exists neither on Phabricator nor on Phragile, you now get an error page.
If the sprint exists on Phabricator but not on Phragile you get a form to fill in the project (and start and end date when not using the sprint extension) which makes it easy to connect existing Phabricator project with Phragile.

PR #145, opened 2015-08-18
Add sprint dates and title to settings dialog

Benestar:
Fixes https://phabricator.wikimedia.org/T108294

<s>Tests missing</s>

### DIFF
--- a/resources/views/sprint/view.blade.php
+++ b/resources/views/sprint/view.blade.php
@@ -163,4 +163,19 @@
 
 @section('optional_scripts')
 	{!! HTML::script('js/sprint_overview.js') !!}
+	{!! HTML::script('js/datepicker.js') !!}
+
+	<script type="text/javascript">
+		var settings = {
+			format: 'yyyy-mm-dd',
+			autoclose: true
+		};
+
+		$('.datepicker.start').datepicker(settings);
+		$('.datepicker.end').datepicker(settings);
+	</script>
+@stop
+
+@section('optional_styles')
+	{!! HTML::style('/css/datepicker.css') !!}
 @stop
--- a/resources/views/sprint/view.blade.php
+++ b/resources/views/sprint/view.blade.php
@@ -163,19 +163,6 @@
 
 @section('optional_scripts')
 	{!! HTML::script('js/sprint_overview.js') !!}
-	{!! HTML::script('js/datepicker.js') !!}
-
-	<script type="text/javascript">
-		var settings = {
-			format: 'yyyy-mm-dd',
-			autoclose: true
-		};
-
-		$('.datepicker.start').datepicker(settings);
-		$('.datepicker.end').datepicker(settings);
-	</script>
 @stop
 
-@section('optional_styles')
-	{!! HTML::style('/css/datepicker.css') !!}
-@stop
+@include('sprint.partials.datepicker_assets')
--- a/resources/views/sprint/view.blade.php
+++ b/resources/views/sprint/view.blade.php
@@ -162,7 +162,7 @@
 @stop
 
 @section('optional_scripts')
+	@parent
+
 	{!! HTML::script('js/sprint_overview.js') !!}
 @stop
-
-@include('sprint.partials.datepicker_assets')
--- a/tests/acceptance/bootstrap/FeatureContext.php
+++ b/tests/acceptance/bootstrap/FeatureContext.php
@@ -178,24 +178,26 @@ class FeatureContext extends MinkContext implements Context, SnippetAcceptingCon
 		Auth::login(User::where('username', $this->params['phabricator_username'])->first()); // this is a bit ugly.
 
 		$project = Project::firstOrCreate(['title' => $projectTitle]);
-		$existingSprint = Sprint::where('title', $sprintTitle)->first();
+		$phabricatorProject = $this->getPhabricatorProjectFromTitle($sprintTitle);
+		$existingSprint = Sprint::where('phid', $phabricatorProject['phid'])->first();
 
-		if (!$existingSprint)
+		if ($existingSprint !== null && !$existingSprint->delete())
 		{
-			$phabricatorProject = $this->getPhabricatorProjectFromTitle($sprintTitle);
-			$newSprint = new Sprint([
-				'title' => $sprintTitle,
-				'project_id' => $project->id,
-				'sprint_start' => '2014-12-01',
-				'sprint_end' => '2014-12-14',
-				'phabricator_id' => $phabricatorProject['id'],
-				'phid' => $phabricatorProject['phid'],
-			]);
+			throw new Exception('Could not delete the existing sprint.');
+		}
 
-			if (!$phabricatorProject || !$newSprint->save())
-			{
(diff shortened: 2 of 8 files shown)


resources/views/sprint/view.blade.php
```diff
@@ -163,4 +163,19 @@
...
+			autoclose: true
+		};
+
+		$('.datepicker.start').datepicker(settings);
+		$('.datepicker.end').datepicker(settings);
+	</script>
+@stop
+
+@section('optional_styles')
+	{!! HTML::style('/css/datepicker.css') !!}
```
Benestar: It's a bit ugly to add those scripts and styles here but somehow they don't work if I put them into settings_form.blade.php. @jakob-WMDE thoughts?

jakobw: I did something similar here. https://github.com/wmde/phragile/pull/141/files#diff-3862b26d031b9f01adf42a25d4299fa4R28

Benestar:
Last commit per https://github.com/laravel/docs/blob/4.2/templates.md#using-a-blade-layout

jakobw:
Do we want to be able to edit the title? It would be confusing if the name is different from the one on Phabricator.

Benestar:
If you change the date of the sprint, I'm pretty sure you also want to fix the title of the sprint as it usually contains the date. The title should be editable on phragile if it changes on phabricator.

jakobw:
The use cases for this if I remember it correctly was a) accidental clicking the wrong date (happened to Tobi) or b) adjusting the end date which was a request from the WMF. I don't think it's very harmful though...

jakobw:
The changes need validations. I can set the title to `""` (empty string) and can also do crazy things with the start and end date which shouldn't happen. It should at least have the same validations as the sprint creations.

Benestar:
Validation is added in the latest commit.

> It would be confusing if the name is different from the one on Phabricator.

If those titles should always be the same, then we should fetch them from phabricator directly instead of storing them in our database (or check for changes to the sprint project's title and update them respectively). The easier solution for now is imo to allow the user to change the title manually.

jakobw:
This isn't working on my local machine. Neither manually nor using the behat tests. If I try to change the sprint start or end I get "The title has already been taken.". I'm assuming the uniqueness validation fails because it thinks you're trying to create a new sprint.

Benestar:
Per http://laravel.com/docs/4.2/validation#rule-unique I have added the sprint's own id to the ignore field of laravels uniqueness constraint.

tests/acceptance/sprint_settings.feature
```diff
@@ -3,7 +3,29 @@ Feature: Sprint Settings
...
+
+  Scenario: Edit sprint title
+    Given I am logged in
+    And a sprint "Sprint 42" exists for the "Wikidata" project
+    When I go to the "Sprint 42" sprint overview
+    And I fill in "title" with "Sprint Foo Bar"
+    And I press "save-sprint-settings"
+    Then I should see "The sprint settings have been updated"
+    And the "title" field should contain "Sprint Foo Bar"
+    And I should see "Wikidata Sprint Foo Bar" on "/projects/wikidata"
```
jakobw: This doesn't work because the sprint we just edited will not in every case be the most recent sprint on "/projects/wikidata". I guess it's fine to just remove this step.

tests/acceptance/sprint_settings.feature
```diff
@@ -11,3 +33,26 @@ Feature: Sprint Settings
...
+    Given I am logged in
+    And a sprint "Sprint 42" exists for the "Wikidata" project
+    And the sprint "Sprint 42" starts on "2014-11-25" and ends on "2014-12-08"
+    When I go to the "Sprint 42" sprint overview
+    And I fill in "sprint_start" with "26.11.2014"
+    And I fill in "sprint_end" with "2014-11-20"
+    And I fill in "title" with ""
+    And I press "save-sprint-settings"
+    Then I should see "The sprint start does not match the format Y-m-d."
+    And I should see "The sprint end must be a date after 2014-11-25."
```
jakobw: It won't complain about the date format and date order at the same time.

tests/acceptance/sprint_settings.feature
```diff
@@ -11,3 +33,26 @@ Feature: Sprint Settings
...
+    And I should see "The title field is required."
+
+  Scenario: Duplicate title
+    Given I am logged in
+    And a sprint "Sprint 42" exists for the "Wikidata" project
+    And a sprint "Sprint Foo Bar" exists for the "Wikidata" project
```
jakobw: Something weird happens here. Since you previously renamed "Sprint 42" to "Sprint Foo Bar", creating a new "Sprint 42" here causes both sprints to point to the same Phabricator project. It would be good to revert the renaming before doing something like this.

Benestar:
I hope I fixed everything. At least the tests now run on my local machine \o/ There was some data in the database without ids that must've caused issues.

jakobw:
Heh. You're right, it doesn't. It should though but I guess that should not be tackled here.

jakobw:
@Benestar I get green tests with #153.
The issue was that even if I manually created the sprint it failed on https://github.com/wmde/phragile/pull/145/files#diff-0503986c398e2465a4b76d249d221d07R65 since the sprint got previously deleted from both Phragile **and Phabricator** in https://github.com/wmde/phragile/pull/145/files#diff-0503986c398e2465a4b76d249d221d07R21

#153 makes sure there is always a Phabricator project when using the sprint exists step and then everything works \o/

tests/acceptance/bootstrap/FeatureContext.php
```diff
@@ -165,9 +165,16 @@ public function theSprintDoesNotExist($title)
...
 	}
 
-	private function getPhabricatorProjectFromTitle($title)
+	private function getOrCreatePhabricatorProjectFromTitle($title)
 	{
-		return App::make('phabricator')->queryProjectByTitle($title);
+		$phabricator =  App::make('phabricator');
+		$project = $phabricator->queryProjectByTitle($title);
+
+		if (!$project) {
```
jakobw: Curly bracket should be on the next line.

Also this would be even more compact but I don't object to the current solution:

    $phabricator = App::make('phabricator');
    return $phabricator->queryProjectByTitle($title) ?: $phabricator->createProject($title, []);

tests/acceptance/bootstrap/FeatureContext.php
```diff
@@ -165,9 +165,16 @@ public function theSprintDoesNotExist($title)
 		}
 	}
 
-	private function getPhabricatorProjectFromTitle($title)
+	private function getOrCreatePhabricatorProjectFromTitle($title)
 	{
-		return App::make('phabricator')->queryProjectByTitle($title);
+		$phabricator =  App::make('phabricator');
```
jakobw: An extra space snuck into the assignment after `=`.

tests/acceptance/bootstrap/FeatureContext.php
Benestar: Gah, different code styles everywhere. Confusing!

Indeed your snippet is even shorter. I will use it :)

jakobw:
Yeah \o/

PR #148, opened 2015-08-19
Fix scenario 'Assignees in Sprint Backlog'.

jakobw:
This fixes a scenario that previously always failed. It contained a hardcoded task ID which is now replaced. The fix introduces some state flying around in `FeatureContext` which isn't great but having the tests pass is more valuable in my opinion.
There is another scenario failing because of this which I will fix in a separate pull request.

Poke @wmde-manicki @Benestar


wmde-manicki:
Yeah, more acceptance tests passing!
And I confirm, the test for "Snapshots are not affected by Phabricator actions" scenario is now complaining about missing step for "A contains task B". But as this tests are not yet being run on CI etc I would be fine with merging this by now.

PR #154, opened 2015-08-31
Fix sprint overviews for large projects

jakobw:
Task on Phab: https://phabricator.wikimedia.org/T110330

There were two different problems with large projects.

1. **Timeouts on task transaction requests**
It is not possible to load all task transactions of large projects at once through Conduit (Phabricator's API). This was solved by fetching transactions in batches of 200. More details: https://phabricator.wikimedia.org/T109036

2. **Memory issues when processing/storing the transactions**
Loading all transactions is solved by loading them in batches, however, processing all the transactions consumes a lot of memory. This can be improved by filtering out irrelevant transactions. For snapshots it'll store all transactions regarding workboard and status changes so that switching between workboard and status mode still works. Unfortunately this still exceeded the memory limit for sprint overviews but ~~can be~~ **that has been** fixed by also removing status transactions in workboard mode and workboard transactions in status mode.

I deployed the changes to a test instance and you can check it out here: http://phragile-test.wmflabs.org/sprints/483

~~Snapshots are still somewhat broken for large projects unfortunately. For some reason they can be created but not read.~~
Snapshots are fixed in #155

### DIFF
--- a/app/Phragile/SettingsAwareTransactionFilter.php
+++ b/app/Phragile/SettingsAwareTransactionFilter.php
@@ -1,6 +1,10 @@
 <?php
 namespace Phragile;
 
+/***
+ * Class for filtering out all transactions but the ones relevant to the task status according to the project settings
+ * @package Phragile
+ */
 class SettingsAwareTransactionFilter extends TransactionFilter {
 	private $workboardMode;
 
--- a/app/Phragile/SettingsAwareTransactionFilter.php
+++ b/app/Phragile/SettingsAwareTransactionFilter.php
@@ -1,0 +1,17 @@
+<?php
+namespace Phragile;
+
+class SettingsAwareTransactionFilter extends TransactionFilter {
+	private $workboardMode;
+
+	public function __construct($workboardMode)
+	{
+		$this->workboardMode = $workboardMode;
+	}
+
+	protected function isRelevantTransaction(array $transaction)
+	{
+		return ($this->workboardMode && $this->isWorkboardTransaction($transaction))
+			|| (!$this->workboardMode && $this->isStatusTransaction($transaction));
+	}
+}
--- a/app/Phragile/SettingsAwareTransactionFilter.php
+++ b/app/Phragile/SettingsAwareTransactionFilter.php
@@ -1,9 +1,8 @@
 <?php
 namespace Phragile;
 
-/***
+/**
  * Class for filtering out all transactions but the ones relevant to the task status according to the project settings
- * @package Phragile
  */
 class SettingsAwareTransactionFilter extends TransactionFilter {
 	private $workboardMode;
--- a/app/Phragile/StatusByWorkboardDispatcher.php
+++ b/app/Phragile/StatusByWorkboardDispatcher.php
@@ -47,16 +47,10 @@ class StatusByWorkboardDispatcher implements StatusDispatcher {
 		});
 	}
 
-	private function getTaskColumnName(array $task)
-	{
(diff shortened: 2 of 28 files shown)


Benestar:
Would you mind splitting this into several small pull requests or at least re-submit #155 against master. That would make reviewing this stuff a lot easier. ;-)

Benestar:
Rebased to get the changes of #160 in here.

app/Phragile/SettingsAwareTransactionFilter.php
```diff
@@ -0,0 +1,21 @@
+<?php
+namespace Phragile;
+
+/***
```
Benestar: I think we use only two starts in this project? Or is this inconsistent anyways?

app/Phragile/SettingsAwareTransactionFilter.php
```diff
@@ -0,0 +1,21 @@
...
+ * Class for filtering out all transactions but the ones relevant to the task status according to the project settings
+ * @package Phragile
```
Benestar: `@package`? What does that do/mean?

app/Phragile/TransactionLoader.php
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	/***
+	 * Loads task transactions from Phabricator in batches of 200
+	 *
+	 * @param $taskIDs
+	 * @param PhabricatorAPI $phabricatorAPI
+	 * @return array
+	 */
+	public function load($taskIDs, PhabricatorAPI $phabricatorAPI)
```
Benestar: I'd say the PhabricatorAPI object is a service that should be passed in via the constructor.

app/Phragile/StatusByWorkboardDispatcher.php
```diff
@@ -46,10 +47,16 @@ private function findCurrentColumn(array $taskTransactions)
 		});
 	}
 
+	private function getTaskColumnName(array $task)
+	{
+		$phid = isset($this->taskColumnPHIDs[$task['id']]) ? $this->taskColumnPHIDs[$task['id']] : null;
+		return $this->columns->getColumnName($phid) ?: 'Backlog';
```
Benestar: Code duplication..

app/Phragile/StatusByWorkboardDispatcher.php
```diff
@@ -46,10 +47,16 @@ private function findCurrentColumn(array $taskTransactions)
...
+	}
+
 	public function isClosed(array $task)
 	{
 		return in_array(
-			$this->columns->getColumnName($this->taskColumnPHIDs[$task['id']]),
+			$this->getTaskColumnName($task),
```
Benestar: Why not just use `getStatus`

Benestar:
This still does two things in a row that could possibly be split but I think if the mentioned comments get resolved/answered we can merge this :-)

Benestar:
> For snapshots it'll store all transactions regarding workboard and status changes so that switching between workboard and status mode still works. Unfortunately this still exceeded the memory limit for sprint overviews but can be fixed by also removing status transactions in workboard mode and workboard transactions in status mode.

Does that mean that with this patch it is still impossible to view a sprint? Or to view a snapshot? Do we need to disable switching the mode for snapshots?

jakobw:
@Benestar It's possible to view sprints and snapshots with this patch :)
Wording was a bit off there. It is actually removing all irrelevant transactions now for live sprint overviews and only keeping the ones for status and workboard transactions on snapshots.

jakobw:
@Benestar It would be awesome if you could have another look at this :)

Benestar:
I just noticed that http://phragile-test.wmflabs.org/sprints/782 looks very strange. But at least it loads something!

PR #155, opened 2015-08-31
Lazy load snapshot data

jakobw:
Snapshots caused memory issues for large projects on sprint overviews since the entire snapshot including archived task and transaction data got loaded for each existing snapshot for the sprint.
This patch solves this by lazy-loading the snapshots' data attribute.

Fixes https://phabricator.wikimedia.org/T110649

### DIFF
--- a/app/Http/Controllers/SprintSnapshotsController.php
+++ b/app/Http/Controllers/SprintSnapshotsController.php
@@ -54,7 +54,7 @@ class SprintSnapshotsController extends Controller {
 
 	private function getSprintDataFactory(SprintSnapshot $snapshot)
 	{
-		$sprintData = json_decode($snapshot->data, true);
+		$sprintData = json_decode($snapshot->getData(), true);
 		return new SprintDataFactory(
 			$snapshot->sprint,
 			$sprintData['tasks'],
--- a/app/Models/Sprint.php
+++ b/app/Models/Sprint.php
@@ -22,7 +22,9 @@ class Sprint extends Eloquent {
 	 */
 	public function sprintSnapshots()
 	{
-		return $this->hasMany('SprintSnapshot')->orderBy('created_at', 'desc');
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/SprintSnapshotsController.php
```diff
@@ -54,7 +54,7 @@ public function delete(SprintSnapshot $snapshot)
 
 	private function getSprintDataFactory(SprintSnapshot $snapshot)
 	{
-		$sprintData = json_decode($snapshot->data, true);
+		$sprintData = json_decode($snapshot->getData(), true);
```
Benestar: Is that really needed here? I think in the SprintSnapshotsController the full snapshot data will be loaded anyways. Am I wrong?

app/Models/Sprint.php
```diff
@@ -22,7 +22,9 @@ public function project()
 	 */
 	public function sprintSnapshots()
 	{
-		return $this->hasMany('SprintSnapshot')->orderBy('created_at', 'desc');
+		return $this->hasMany('SprintSnapshot')
+		            ->select(['id', 'sprint_id', 'created_at', 'total_points'])
+		            ->orderBy('created_at', 'desc');
```
Benestar: Makes totally sense to not load the actual data here. Should maybe give a hint in the method comment that this will return only parts of the SprintSnapshot objects.

app/Http/Controllers/SprintSnapshotsController.php
jakobw: You're right that it would not be needed for routes that use model binding using `SprintSnapshot::find` to get the snapshot, however, `SprintsController` is delegating to `SprintSnapshotsController` in two places and it is needed for these cases.

Benestar: Ah I see. Then it makes sense to have this here :-)

PR #156, opened 2015-09-03
Number of tasks in snapshots

jakobw:
This fixes the issue that when switching `ignore_estimates` on or off after taking a snapshot, the scope line breaks since there is no differentiation between story points and task count with `ignore_estimates` enabled. To resolve that, snapshots now contain both the task count as well as the total number of story points in the sprint.

Fixes https://phabricator.wikimedia.org/T110892

### DIFF
--- a/app/Models/Sprint.php
+++ b/app/Models/Sprint.php
@@ -101,6 +101,7 @@ class Sprint extends Eloquent {
 			'sprint_id' => $this->id,
 			'data' => json_encode($this->fetchSnapshotData($tasks)),
 			'total_points' => $this->calculateTotalPoints($tasks),
+			'task_count' => count($tasks),
 		]);
 	}
 
--- a/app/Models/Sprint.php
+++ b/app/Models/Sprint.php
@@ -111,7 +111,7 @@ class Sprint extends Eloquent {
 			$tasks,
 			function($sum, $task)
 			{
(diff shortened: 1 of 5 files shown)


wmde-manicki:
Looks fine and works good, very nice! But would need a rebase to get merged.
My only concern how would look the scope line for snapshots that had been taken before this change. Isn't it going to be flat line at zero? I know I could find out by trying this out, and I did it but I am not sure that it is what I am getting. Have you checked such cases?

jakobw:
@wmde-manicki Rebased. I tried it out and aside from breaking the scope line values of burnup charts I nothing bad happened. I guess it's fine since I don't think this `ignore_estimates` feature is being used much as of now.

tobijat:
@jakobw actually it is used in the Wikidata project quite often (in the latest 3 sprints). Is it possible to not break the burnup and scopeline charts for these?

wmde-manicki:
@jakobw: great, it also seemed rather fine when I was trying this out, although my test phabricator projects are rather limited.
But given @tobijat's comment I don't dare to click "Merge" at this point.

jakobw:
@wmde-manicki I talked to @tobijat and I'll just run this little snippet from the laravel console to migrate the affected sprints:

    Sprint::where('ignore_estimates', true)->get()->each(function($sprint)
    {
      $sprint->sprintSnapshots->each(function($snapshot)
      {
        $snapshot->task_count = $snapshot->total_points;
        $snapshot->save();
      });
    });

wmde-manicki:
Yeah!

PR #157, opened 2015-09-03
Fix labels in burn up/down chart.

jakobw:
Fixes https://phabricator.wikimedia.org/T110669


wmde-manicki:
I have made a task with 1000 story points and give this change a try (see below). It does look much better than it would before (than even 200 label gets cropped) but it still does not seem to fit.
But this is kind of hard to resolve. If the labels got even shrunken even more, so 1000 fits well, it is possible that the label still gets cropped if the total number of points grows to 10k, and so on.

phragile-1000storypoints-labels

jakobw:
@wmde-manicki Should be fixed now. I was being smart and didn't actually add a task with over 1000 but just edited it manually in the Chrome debugger. I didn't expect the comma :|

wmde-manicki:
Looks good now!

PR #159, opened 2015-09-07
Root images displayed in index page

addshore:
If phragile is installed in a directory on a domain
for example localhsot/ph these images do not currently
get displayed correctly!


Benestar:
I believe this won't work on subpages...

addshore:
Ahh, indeed, the header is included on subpages too, duh!
I guess we need to get the url of phragile from somewhere
Or say installing in directories is not supported? ;)

PR #162, opened 2015-10-27
Use env() instead of $_ENV[].

jakobw:
Using `env()` will make it possible to run Phragile without Apache/nginx by using `php artisan serve`.

### DIFF
--- a/app/Http/Controllers/SessionsController.php
+++ b/app/Http/Controllers/SessionsController.php
@@ -7,7 +7,7 @@ class SessionsController extends Controller {
 
 	public function __construct()
 	{
-		$this->phabricatorOAuth = new PhabricatorOAuth($_ENV['PHABRICATOR_URL']);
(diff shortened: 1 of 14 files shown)


addshore:
Are you sure you dont mean this?
http://php.net/manual/en/function.getenv.php

jakobw:
Yup, I'm sure. Not 100% sure why, but it seems that environment variables from the config file aren't available through `getenv` or `$_ENV` when using `php artisan serve` as fooishbar found out by accident https://phabricator.wikimedia.org/T100190#1728686 
`env()` is a Laravel helper function http://laravel.com/docs/5.0/helpers#miscellaneous

dereckson:
Indeed, `env()` is the way Laravel reads by default environment variables.

This allows to specify in DotEnv files (.env) special values like `(empty)`, which will be interpreted as an empty string (`''`)

PR #163, opened 2016-01-06
Show burnup or burndown depending on the url fragment.

jakobw:
Task: https://phabricator.wikimedia.org/T119186

Putting #burnup at the end of a Phragile sprint URL will now show the burnup chart right away.


tobijat:
It was not written directly in the task, but I think it should also add the #burnup or #burndown to the URL when I click the tabs. So I'm able to reload the page and see the same tab immediately.

wmde-manicki:
Just tested it locally and apart from @tobijat said it looks and works good.

jakobw:
Updated. In case you're wondering what's up with the weird `#!`syntax - it keeps the browser from jumping to the burndown chart when #burndown is selected.

wmde-manicki:
Works for me! I'd merge this unless @tobijat has some further remarks.

WMDE-Fisch:
+1 as well

wmde-manicki:
@tobijat just said he has no further objections

PR #164, opened 2016-01-07
Change the project a sprint is attached to

wmde-manicki:
Task: https://phabricator.wikimedia.org/T119104

https://github.com/wmde/phragile/commit/0bea94fe8797d841d0b3e075db2ab0037ebea8b4 has been added for cases when the sprint is the current sprint of the project, ie. accessed through the `/projects/foo-project` path. In such cases going "back" after saving the changed settings would show different sprint of the "old" project, instead of the just-updated sprint. I'd find that confusing.


wmde-manicki:
OK, I have fixed my problems with making Behat click on the project "drop-down" list but as Mink dirvers (BrowserKit and Goutte) do not evaluate JS code, it is impossible to actually test this feature, as it involves manipulation of hidden field's value with JavaScript.
In order to have Behat tests for parts of Phragile that use JS some other driver should probably be used, see: http://mink.behat.org/en/latest/guides/drivers.html#driver-feature-support
This could actually might be desired to have more acceptance tests for other features that have not been currently covered by tests due to the lack of JS "support" in the driver.
This change would probably mean some more complex re-structure of Phragile's acceptance tests (e.g. there is no need to change drivers for currently existing tests). Any thoughts on what and how should we do here @jakob-WMDE and our test maestro @WMDE-Fisch ?

As there is no longer remaining work from my perspective, I am removing the `[WIP]` tag from PR's title. If adding code related to new Mink drivers etc is requested here, I'll do it.
Work on the actual feature is finished as for now.

jakobw:
+1 in general for a Behat driver that works with JS.

In this particular case I don't really see the benefit of the fancy Bootstrap dropdown over a normal `select` box. Aside from the fact that it may not look as pretty it solves all our problems, doesn't it? In fact there is already a projects select box used on the connect_sprint page. https://github.com/wmde/phragile/blob/master/resources/views/sprint/connect.blade.php#L24

wmde-manicki:
Ugh, I looked at other "drop-downs" in the app and they are using this button+ul combo, but I've obviously missed that one.
Then it is of course better to do it with `select`. I'll re-do this patch

wmde-manicki:
Changed the whole patch. Thanks @jakobw for enlightening comment. I found a way to make the dropdown look almost as the "fancy bootstrap" one.

jakobw:
Works perfectly and the dropdown looks cool! :D

PR #171, opened 2016-01-12
Ignore workboard columns

jakobw:
Task: https://phabricator.wikimedia.org/T113716

Ignoring columns for snapshots got a little bit tricky since tasks did not know their status at the time the snapshot was created until now. I created a `StatusDispatcherFactory` to get the task status depending on a sprint's `workboard_mode` in both the `SprintDataFactory` and in during snapshot creation.

### DIFF
--- a/app/Http/Controllers/ProjectsController.php
+++ b/app/Http/Controllers/ProjectsController.php
@@ -43,7 +43,7 @@ class ProjectsController extends Controller {
 
 	public function updateSettings(Project $project)
 	{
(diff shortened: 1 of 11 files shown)


wmde-manicki:
One tiny remark: Shouldn't the "Ignored workboard columns:" option hide similar to the "Closed status columns:" when users unchecks the "Use workboards instead of statuses"?

jakobw:
@wmde-manicki Good point! Updated to include all tasks in snapshots even if they're ignored and also hide the setting unless the project is in workboard mode.

PR #173, opened 2016-01-13
Default column

jakobw:
Task: https://phabricator.wikimedia.org/T118945

* added the `default_column` field to the DB, model, settings form and accepted parameters of the controller + a little acceptance test for the setting
* decided to pass the whole `Sprint` object to the `StatusByWorkboardDispatcher` to avoid having to pass even more of the `Sprint` fields to it. This caused most of the changes in the `TaskListTest`.
* replaced the hardcoded `'Backlog'` with the new setting
* added a unit test for the `TaskList`

### DIFF
--- a/app/Http/Controllers/ProjectsController.php
+++ b/app/Http/Controllers/ProjectsController.php
@@ -43,7 +43,7 @@ class ProjectsController extends Controller {
 
 	public function updateSettings(Project $project)
 	{
-		$project->update(Input::only('closed_statuses', 'workboard_mode', 'ignored_columns', 'backlog_column'));
+		$project->update(Input::only('closed_statuses', 'workboard_mode', 'ignored_columns', 'default_column'));
 
 		Flash::success('The project settings have been updated');
 		return Redirect::back();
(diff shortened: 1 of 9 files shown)


WMDE-Fisch:
Looks good and I would merge it, but: I don't think that the term "Default Column" is specific enough. I think it's not clear here what the setting is used for.

What we _mean_ is something like "Default Backlog Column" and since there is no other column than the default one (atm) we could also rename it to "Backlog column".

And this should also be done with the identifiers in the code imo.

wmde-manicki:
+1
Looks fine, and seems to be working as intended. I agree with what @WMDE-Fisch says, it might be more clear to talk about "default backlog" column, as there is something like a "default" column in phabricator too (although I cannot find a way to change it now), which I guess might not always mean the same thing as in phragile.
Given @WMDE-Fisch has gone forward and changed things he found, this could be actually merged. I'd still rather wait for @jakobw to address the comments.

jakobw:
Hm, I'm not 100% happy with `default_backlog_column`. Maybe `default_workboard_column`? We have this other thing called `ignored_columns` flying around. Should this be renamed too?

jakobw:
Updated!

PR #175, opened 2016-01-14
use token based authentication instead of certificates

WMDE-Fisch:
Including fallbacks for certificate based bot configuration and for users that did not set an api token yet.

see https://phabricator.wikimedia.org/T123514

This can only be used with an updated version of libphutil ( including the same minor modifications as in the current libphutil ). See Phabricator ticket.


WMDE-Fisch:
https://github.com/wmde/phragile/pull/178 should be merged first

resources/views/layouts/partials/conduit_api_token_form.blade.php
```diff
@@ -3,26 +3,25 @@
...
-							env('PHABRICATOR_URL') . '/settings/panel/conduit/',
-							'Phabricator &gt; Settings &gt; Conduit Certificate'
+							env('PHABRICATOR_URL') . '/settings/panel/apitokens/',
+							'Phabricator &gt; Settings &gt; Conduit API Tokens'
 						) !!}
 					</p>
 					<p>
 						{!! Form::textarea(
-							'conduit_certificate',
+							'conduit_api_token',
```
jakobw: Small thing: API tokens are much shorter so this would look nicer with a text input field instead of a textarea :)
Also, we get to remove the `iShouldSeeMyConduitAPITokenInATextArea` method and just use the built-in `Then the "" field should contain ""` check.

jakobw:
I was able to test this and aside from the tiny comment everything looks good and works really well. Awesome!
I'll open a ticket about the questions concerning Conduit certificate support.

WMDE-Fisch:
updated the PR to use an input field instead of a textarea

PR #177, opened 2016-01-15
added defaults and change field to VARCHAR

WMDE-Fisch:
in MYSQL text cant have a default value
since Phabricator itself uses VARCHAR for this it should be safe to use it

also added another default empty value for string

see https://phabricator.wikimedia.org/T117323
and https://phabricator.wikimedia.org/T117319

### DIFF
--- a/database/migrations/2016_01_15_110323_add_default_values.php
+++ b/database/migrations/2016_01_15_110323_add_default_values.php
@@ -1,0 +1,43 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+
+class AddDefaultValues extends Migration {
+
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		Schema::table('users', function($table)
+		{
+			$table->string('conduit_certificate')->default('')->change();
+		});
+		Schema::table('projects', function($table)
+		{
+			$table->string('closed_statuses')->default('')->change();
+			$table->string('ignored_columns')->default('')->change();
+		});
+	}
+
+	/**
+	 * Reverse the migrations.
+	 *
+	 * @return void
+	 */
+	public function down()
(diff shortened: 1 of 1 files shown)


wmde-manicki:
+1 as seems to fix at least a problem reported in https://phabricator.wikimedia.org/T117323. I am not merging it now as I am guessing that @jakobw might like to check what up with this PR and https://github.com/wmde/phragile/issues/178 fixing the same things :)

But to add strangess here. When I tried to rollback migration added in the PR, I got a nice errr message saying among others that `[PhutilMissingSymbolException]                                                                                                                                                                                   
  Failed to load class or interface 'AddDefaultValues': the class or interface 'AddDefaultValues' is not defined in the library map for any loaded phutil library. If this symbol was recently added or moved, yo  
  ur library map may be out of date. You can rebuild the map by running 'arc liberate'. For more information, see: http://www.phabricator.com/docs/phabricator/article/libphutil_Libraries_User_Guide.html` I am guessing that might be something related to that libphutil update (which I haven't done locally yet) or something like this.

WMDE-Fisch:
Related to the Error: This is is normal when switching between branches with migrations. When you checkout a branch run migrate, when you want to checkout another branch do rollback. If you are already on the other branch and try to rollback the migration class needed for the rollback might not be found since it is in the other branch.

It gets even more complicated when you rename a migration/change class inside a branch before rollback.

jakobw:
@wmde-manicki I'm guessing that it was really just because you switched branches and the rollback command tried to roll back something from a class that wasn't there anymore in the other branch. I don't think it has something to do with libphutil. libphutil somehow messes with exceptions which is why you don't see a normal PHP error.

wmde-manicki:
Yeah I suspect that too, although I tried to be careful and not jump between branches. But apparently I did it, though :)

PR #178, opened 2016-01-15
updating libraries used in composer.lock

WMDE-Fisch:
fixes several problems I and other encountered with the current libraries uses
is also needed for updated libphutil

also fixes https://phabricator.wikimedia.org/T117324

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,27 +1,28 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
     "hash": "6639061b3465a3ee8dcdf03c19fb0e51",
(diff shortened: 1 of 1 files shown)


wmde-manicki:
+1 for the change
It is unclear for me how does this fixes the problem mentioned in the phab ticket, though. Some DB-stuff got changed and now empty string defult to zero?
Using dependencies defined on the current master I couldn't reproduced the error described in that ticket too, so it's all strange. But I'm totally fine with the fact that this change fixes that problem :)

wmde-manicki:
Breaking news: I've managed to replicate the problem on master

WMDE-Fisch:
Strange now I cant replicated it anymore :-)

wmde-manicki:
So the problem with `Integrity constraint violation: 1048 Column 'ignore_estimates' cannot be null ` appeared after setting `sql_mode=STRICT_ALL_TABLES` in mysql config. It looks clearly like a problem with passing null to non-null-allowed column. Apparently some change in some dependency fixes something related to that.
Strangely enough, problems reported in https://phabricator.wikimedia.org/T117323 do not appear any more even using `STRICT_ALL_TABLES` when I have updated dependencies but not applied patches from: https://github.com/wmde/phragile/pull/177 :)

WMDE-Fisch:
Hmm super strange. But default values are good anyway I would say :-)

PR #179, opened 2016-01-19
Bump version of the custom libphutil library

wmde-manicki:
In order to use Phabricator API conduit token instead of conduit certificate, "hacked" libphutil should be updated.
As the custom version of the library has no "real" releases, composer does not notice when it changes.
This is probably not the most elegant way to solve the problem but bumping the version number makes composer to update the library if it has been already installed.
Actually, this might be the only way to update that kind of "package", according to note in https://getcomposer.org/doc/05-repositories.md#package-2

As also mentioned in the Composer documentation, I am removing the `source` definition as it has no use for this dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,12 +14,7 @@
 					"url": "https://github.com/jakobw/libphutil/archive/master.zip",
 					"type": "zip"
 				},
-				"source": {
-					"url": "https://github.com/jakobw/libphutil.git",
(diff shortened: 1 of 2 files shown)


WMDE-Fisch:
The other question is if we want to provide different versions of "our" libphutil when I change from this "version" to the "version" before he will end up loading the same files.

PR #180, opened 2016-01-20
added Background to some features to reduce repetition

WMDE-Fisch:


jakobw:
Looks very nice!
I wonder if there are other behat features (maybe [hooks](http://docs.behat.org/en/v2.5/guides/3.hooks.html)?) that we aren't using yet that could help us with the clean-up.

wmde-manicki:
+1 Looks good! Three tests fail for me but they also fail on master, so this does not seem to be related to this change.
Those hooks indeed look like something that could be nicely included in the tests here.

WMDE-Fisch:
Nice hint with the hooks. If I see something where this makes sense I would create a PR to discuss it. So plz merge if you feel these steps are already useful :-)

PR #181, opened 2016-01-21
Drop not needed before script in TravisCI builds

wmde-manicki:
Same command being run in the `install` step as well. I believe once is enough.


wmde-manicki:
Same command being run in the `install` step as well. I believe once is enough.

PR #182, opened 2016-01-21
changed menu to allow sprint creation and project settings in project view

WMDE-Fisch:
This is not only a nice little improvement but also helpful for the tests.
Now there is a consistent way of creating new sprints for a project independent of if is has already a sprint or not.

PR #183, opened 2016-01-21
test independence

WMDE-Fisch:
Includes https://github.com/wmde/phragile/pull/182 what should be merged first.

By adding 

	/** @BeforeScenario */
	public function before($event)
	{
		exec('php artisan migrate:refresh');
	}

to your `FeatureContext.php` the Phragile DB should be reset before each _scenario_.  Running this test with this setting show, that each test could run independent of the others on a fresh Phragile installation. - Of cause the Phabricator DB is not cleaned what leaves a small amount of uncertainty. But at least it is a first step.

### DIFF
--- a/tests/acceptance/bootstrap/FeatureContext.php
+++ b/tests/acceptance/bootstrap/FeatureContext.php
@@ -177,8 +177,6 @@ class FeatureContext extends MinkContext implements Context, SnippetAcceptingCon
 	 */
 	public function aSprintExistsForTheProject($sprintTitle, $projectTitle)
 	{
-		Auth::login(User::where('username', $this->params['phabricator_username'])->first()); // this is a bit ugly.
-
 		$project = Project::firstOrCreate(['title' => $projectTitle]);
 		$phabricatorProject = $this->getOrCreatePhabricatorProjectFromTitle($sprintTitle);
 		$existingSprint = Sprint::where('phid', $phabricatorProject['phid'])->first();
--- a/tests/acceptance/bootstrap/FeatureContext.php
+++ b/tests/acceptance/bootstrap/FeatureContext.php
@@ -346,6 +346,15 @@ class FeatureContext extends MinkContext implements Context, SnippetAcceptingCon
 	}
 
 	/**
+	 * @Given I copied the :sprint Phabricator ID from Phabricator
+	 */
+	public function iCopiedThePhabricatorIdFromPhabricator($sprint)
+	{
+		$phabricatorProject = $this->getOrCreatePhabricatorProjectFromTitle($sprint);
+		$this->phabricatorProjectID = $phabricatorProject['id'];
+	}
+
+	/**
 	 * @Given I copied the :project :sprint Phabricator ID
 	 */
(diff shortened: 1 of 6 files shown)


wmde-manicki:
I am only wondering (not saying what @WMDE-Fisch suggested is not a good way to go) how "slow" is rolling back and running all migrations before each test, compared with running some commands that would simply create and empty DB instance. I guess the latter might not be significantly faster given the current DB structure and amount of migrations. Do you have any clue about this @WMDE-Fisch @jakobw ?

jakobw:
I would prefer not to merge this unless we have some useful seed data that we can use for manual testing. I often needed manually create sprints to try out how things look for different numbers of tasks/story points/dates. Having that wiped out every time tests are run would be annoying.

wmde-manicki:
+1 what @jakobw says

jakobw:
Another approach could be to make acceptance tests use their own database. I'm pretty sure I saw somewhere how it could be done using a separate `.env` file for tests and setting the `APP_ENV` accordingly.

tests/acceptance/bootstrap/FeatureContext.php
```diff
@@ -345,6 +344,15 @@ public function aSprintExistsForTheProjectInPhabricatorButNotInPhragile($sprintT
...
+	public function iCopiedThePhabricatorIdFromPhabricator($sprint)
+	{
+		$phabricatorProject = $this->getOrCreatePhabricatorProjectFromTitle($sprint);
+		$this->phabricatorProjectID = $phabricatorProject['id'];
+	}
+
+	/**
 	 * @Given I copied the :project :sprint Phabricator ID
 	 */
 	public function iCopiedThePhabricatorId($project, $sprint)
```
wmde-manicki: doesn't adding that `iCopiedThePhabricatorIdFromPhabricator` method make this `iCopiedThePhabricatorId` redundant?
It seems like the latter has only been used in the `connect_existing_project.feature` and has been replaced by the new method.

WMDE-Fisch:
You do realize, that the code snippet above is _not_ part of the PR. It is just a way to verify, that the current master will totally fail when trying to run individual test in a fresh environment, and that the changes in the PR fix that. 

Leaving the code above they still will create a lot of test data for manual testing :-)

jakobw:
Ya, I just realized that too. Sorry, I panicked in fear of losing all my precious test sprints :D

wmde-manicki:
I did realize that :)
I agree with @jakobw that it wouldn't be a best solution though (even if the performance question I asked above is not that important in our case). I'd opt for having a separate test DB. But it is not part of this PR, true that, so let's stop here.

tests/acceptance/bootstrap/FeatureContext.php
WMDE-Fisch: true thanks for seeing this - fixed :-)

wmde-manicki:
+1 All tests pass for me.
I am not sure whether renaming a test sprint from "Wikidata Sprint 42" to "Sprint 42" is some drastic change. I guess not.

PR #184, opened 2016-01-21
add base path to images and top logo link

WMDE-Fisch:
see https://phabricator.wikimedia.org/T111673

PR #185, opened 2016-01-22
add hint text and changed titles for new sprint

WMDE-Fisch:
Includes https://github.com/wmde/phragile/pull/184 that should be merged first.

- added a hint text explaining the possibility to add existing Phabricator projects as sprints.
- changed the label for the title field to "Title / ID".
- also renamed the page to "Add sprint" instead of "Creat sprint". ( Also thought about "Add new sprint" ). I am open for suggestions here.


jakobw:
Nice, +1 for "add sprint" instead of "new sprint".

I think it would be clearer if the hint would be directly above the title field and am not 100% sure whether the "/ ID" part could be confusing. I would leave the label as just "Title:" and then below that put a hint along the lines of "You can connect an existing Phabricator project with Phragile by entering its title or Phabricator ID in the field below."
Maybe @tobijat wants to have a look at this!

wmde-manicki:
+1 for having a generic "add sprint"
+1 for leaving out "/ ID" part
And I also think a hint just below the title field would be even more user-friendly.
But as @jakobw, let's have world's most active Phragile user @tobijat comment on this.

WMDE-Fisch:
changed and also addressed comments by @tobijat in https://phabricator.wikimedia.org/T119187
maybe its to much now - opinions ?

wmde-manicki:
hmm, how would it sound to you:
 - not have any explanation below the "Add sprint" header (one that is now saying "Create a new sprint<sup>1</sup> or connect to an existing one on Phabricator<sup>2</sup>.")
 - below the title field have a hint "Connect to an existing Phabricator project by using its project title or project ID as a title.", and as a second paragraph/sentence/whatever: " If the provided title does not exist on Phabricator yet, a new project with that title will be created there"

jakobw:
I agree with what @wmde-manicki said. Maybe a little * on the Title label and the explanation should go below the field.

WMDE-Fisch:
.oO( next time we should sit together with Firebug and figure it out ;-) )

jakobw:
Soooo the mini text change I mentioned earlier would be
```
Connect to an existing Phabricator project by entering its title or ID as the title.<br/>
&nbsp;If no project with the provided title exists on Phabricator yet, a new project with that title will be created.
```

wmde-manicki:
It might be getting ridiculous to spend so much time on such a tiny thing but I'd some more comments (sorry!)

I have talked with WMDE UX Maestro about something else but we ended up trying out some other variants of aligning the description text, and he suggested few things:
 - The star could be removed from next to the "Title", as would be still clear for the user that remark is about the title, given the next remark is also taken into account:
 - to make it more evident that the remark is about the "Title" field, the submit button should be moved a bit down. Actually the form looks a bit unclear as it seems that spaces between fields are even a bit bigger than the space between the last field and the button (it might not be true if pixels are counted but still it would be visually better if the button is separated a bit more)
 - Henning also suggested to drop colons in all field labels, as they're not really needed when the field are aligned below the labels. That could be also changed in the Sprint Settings window (same form alignment), I'll make a separate patch for that. But colons should stay in the Project Settings window where field are next the labels, not below,
 - font size of the mini text could be made a bit smaller, so it makes it even more clear that this is the description linking to the title field.

Here is the screenshot of the draft version we came up with using firebug (just assume there're no colons in the sprint date field labels)
screenshot from 2016-01-28 12 09 04

What do you guys think @jakobw @WMDE-Fisch ?
I could amend @WMDE-Fisch's change, so it looks like the one Henning suggested and merge it right away, so we could finally close this task :)

jakobw:
+9000 for all the suggestions

wmde-manicki:
@jakobw so I ended up adding a new commit to he PR that does the things Henning suggested (would be). I still might join it into one with the previous commit if preferred.

PR #186, opened 2016-01-22
adding a stats page

WMDE-Fisch:
see https://phabricator.wikimedia.org/T119949
Can be reached via "Statistics" link in menue ( only admins should see this ).
The test obviously assumes a clean instance on each run. 
Could also be change to more flexible numbers and is open for discussion.

### DIFF
--- a/app/Http/Controllers/StatsController.php
+++ b/app/Http/Controllers/StatsController.php
@@ -1,0 +1,19 @@
+<?php
+
+class StatsController extends Controller {
+
+	public function index()
+	{
+		return View::make('stats.index')->with([
+			'projectCount' => Project::count(),
+			'sprintCount' => Sprint::count(),
+			'sprintsPerProject' => $this->getSprintsPerProject(),
+		]);
+	}
+
+	private function getSprintsPerProject() {
+		return Project::all()->map(function($project) {
+			return [$project->title, $project->sprints->count()];
+		});
+	}
+}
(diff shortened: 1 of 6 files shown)


wmde-manicki:
Stats are indeed generated. I am not sure how this stats page should actually look like. Putting sprints per project counts into simple table might make it more readable.

Also, I'd rather add tests which is clicking the link after the link has been actually added (ie. while at https://github.com/wmde/phragile/commit/43fb797c0e5e4f3c55ddd349bb2ec1c78d81ea3c tests will always fail)

WMDE-Fisch:
> Putting sprints per project counts into simple table might make it more readable. 

Did it as simple as possible, but your right, a table would brush up things.

> Also, I'd rather add tests which is clicking the link after the link has been actually added (ie. while at 43fb797 tests will always fail)

I TTDed here. But yeah it should go in one commit so that test are not failing at any commit.

wmde-manicki:
> I TTDed here. But yeah it should go in one commit so that test are not failing at any commit.

Did you also wear this fancy three-visor cap? Uncle Bob would be proud!
Yeah, I guessed that was the case. It is indeed good way to work on things like that, just feels a bit odd not to have tests together with things they test. But as they're just two consecutive commits, we'll live with that, I guess.

jakobw:
Wohoo for working stats! I agree though that they'd look even better with a table. Also, I would personally prefer it if the tests would work in a "dirty" environment with preexisting sprints and projects.

WMDE-Fisch:
> Also, I would personally prefer it if the tests would work in a "dirty" environment with preexisting sprints and projects.

Yeah already thought that. Will change the tests as well.

WMDE-Fisch:
issues addressed

jakobw:
Yay, all green and tablelized!

PR #187, opened 2016-01-26
Fix failing tests for first authorization.

jakobw:
All behat tests requiring login currently fail if the user has never authorized Phragile OAuth before. This can be tested by revoking Phragile on `[PHABRICATOR]/settings/panel/oauthorizations/`

This should patch fix it!


WMDE-Fisch:
Neat :-)

PR #191, opened 2016-01-27
Fix buggy label highlighting on the x-axis.

jakobw:
Task: https://phabricator.wikimedia.org/T124159

Fixed by adding a class to each tick and then highlighting that class instead of highlighting by index which broke whenever there were more data points than ticks on the x axis.


wmde-manicki:
Looked good on the year-long sprint but then I put on some shorter sprint and it looks like it is now highlighting the day before the day which is hovered in the chart. Please have a look at the screencap below to see what I mean:
phragile-x-axis-hilite-bug

jakobw:
Oops, well spotted. Fixed!

wmde-manicki:
Looks good now!

PR #193, opened 2016-01-28
Switch to use libphutil fork from the wmde namespace

wmde-manicki:
Task: https://phabricator.wikimedia.org/T124037

This is switching to use WMDE's fork: https://github.com/wmde/libphutil.
It uses the same Jakob's commit as https://github.com/jakobw/libphutil did.
Only difference is that WMDE is "more up-to-date" with the upstream (two more commits that they added a week ago). This should not affect Phragile but I think it is worth to have this checked by someone else than only me before we really switch the repo.

Version bump in `composer.*` is needed to have wmde version installed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,10 +11,10 @@
 			"package": {
 				"name": "libphutil",
 				"dist": {
-					"url": "https://github.com/jakobw/libphutil/archive/master.zip",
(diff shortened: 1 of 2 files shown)


jakobw:
Looks good and works perfectly for me!

PR #194, opened 2016-01-29
Show an error when the project cannot be found.

jakobw:
Task: https://phabricator.wikimedia.org/T124034
Going to `/projects/not-a-project` now shows an error and lets the user select a different project from a dropdown with all projects.
While implementing I was questioning whether a dedicated page for this is needed or if we should just redirect to `/` with (or even without) a flash message. Thoughts?

### DIFF
--- a/app/Http/Controllers/ProjectsController.php
+++ b/app/Http/Controllers/ProjectsController.php
@@ -4,10 +4,20 @@ class ProjectsController extends Controller {
 
 	public function show(Project $project)
 	{
+		if (!$project->exists)
+		{
+			return $this->projectNotFound();
+		}
+
 		$currentSprint = $project->currentSprint();
 
 		return $currentSprint ? App::make('SprintsController')->show($currentSprint)
 		                      : View::make('project.view', compact('project'));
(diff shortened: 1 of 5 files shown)


WMDE-Fisch:
> While implementing I was questioning whether a dedicated page for this is needed or if we should just redirect to / with (or even without) a flash message. Thoughts?

It might be nice to have something like "This project does not exist. Do you want to create it?" With link to the (pre filled) project create mask.

WMDE-Fisch:
So in that case dedicated page: Yes :-)

jakobw:
To be honest I don't see much value in having a project creation form on that page. There are only a handful of Phragile admins and those know better than to type an incorrect URL in order to create a project.

WMDE-Fisch:
Hmmm true - have not thought about that. But since we have a page for unkown sprints there should also be a page for unknown projects. So +1

wmde-manicki:
I think current solution (showing an error message and the project selection list) is just fine.
I would say that redirecting to the main page with a flash message would still be fine.

PR #196, opened 2016-02-01
Upgrade to Laravel 5.1

WMDE-Fisch:
A db dump / restore lib for Laravel I wanted to use needs version 5.1 installed. And since upgrade seems to be just fine this patch will do everything needed.

In the new version `php artisan clear-compiled` (automatically run after composer install) needs a directory `bootstrap/cache` to be present. The empty .gitignore file takes care, that git creates that directory on checkout.


WMDE-Fisch:
will try upgrading too 5.2

PR #197, opened 2016-02-01
Use forms instead of links for deleting a sprint, and creating and deleting a sprint snapshot

wmde-manicki:
Task: https://phabricator.wikimedia.org/T109433


resources/views/sprint/view.blade.php
```diff
@@ -29,16 +29,15 @@
...
-					['sprint' => $sprint->phabricator_id],
-					[
-						'class' => 'btn btn-danger btn-lg glyphicon glyphicon-remove',
-						'title' => 'Delete sprint',
-						'onclick' => 'return confirm("Delete this sprint?")'
-					]
-				) !!}
+				{!! Form::open(['route' => ['delete_sprint_path', $sprint->phabricator_id], 'method' => 'DELETE']) !!}
+					{!! Form::submit(
+						'&#xe014;', [
```
Benestar: What's `&#xe014;`?

wmde-manicki: When using the Glyphicons Halflings that prints the "remove" symbol. Previously this has been done by bootstrap css (styling `:before` of links) but this cannot be done the same way for submit buttons.
Your comment reminded me that I forgot that it makes no use to add `glyphicon-remove` class to new delete buttons, thanks!

Benestar: Ok, makes sense. Maybe you can add a comment what that encoded character means for glyphicons? That would make it easier to understand what's going on here

PR #199, opened 2016-02-01
upgrade to Laravel 5.1

WMDE-Fisch:
Recommended for backup library usage. In the new version php artisan clear-compiled (automatically run after composer install) needs a directory bootstrap/cache to be present.

### DIFF
--- a/bootstrap/cache/.gitignore
+++ b/bootstrap/cache/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
 	],
 	"require": {
 		"php": ">=5.4.0",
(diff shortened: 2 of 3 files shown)


wmde-manicki:
+1
Just out of curiosity: what about updating to 5.2 you mentioned in https://github.com/wmde/phragile/pull/196#issuecomment-177942482 ? No way or no need to update now?

wmde-manicki:
Scrutinizer's fail seems foobar but I couldn't re-run it.

WMDE-Fisch:
> Just out of curiosity: what about updating to 5.2 you mentioned in #196 (comment) ? No way or no need to update now?

The update to 5.2 seemed more problematic giving me some errors and was not really necessary right now.

PR #200, opened 2016-02-01
added lavarel backup and restore mechanism

WMDE-Fisch:
This will add a library to easily backup and restore the the Phragile database via artisan.

depends on https://github.com/wmde/phragile/pull/199


wmde-manicki:
Err, that's what I meant saying Travis hhvm build is somewhat unstable. It failed but now passed when just restarted. Created https://github.com/wmde/phragile/pull/201 in case it gets worse.

+1 for the change, not really having a clue about those Laravel backup things.

One question: `config/backup-manager.php` is based on some kind of template, right? Looking at it I guess we are going to use `local` (also given examples you gave in https://phabricator.wikimedia.org/T124035#1986494). Couldn't those things that are unlikely to be used (FTP and the evil cloud things) be removed from it? Not that is critical or so, just out of curiosity.

WMDE-Fisch:
> ). Couldn't those things that are unlikely to be used (FTP and the evil cloud things) be removed from it?

good point :-) - done

jakobw:
It would be so cool if the same could be done for the Phabricator databases!
But still +1, needs rebase though

WMDE-Fisch:
> needs rebase though
done

PR #202, opened 2016-02-02
Randomize behat tests.

jakobw:
Good news
* we can haz randomized behat tests.

Bad news
* ~~our tests appear to be failing when ran in random order~~
* random order is an 3.1RC1 feature even though it was introduced in 2014

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
 		"libphutil": "1.2"
 	},
 	"require-dev": {
-		"behat/behat": "~3.0",
+		"behat/behat": "v3.1.0rc1",
 		"behat/mink": "~1.6",
 		"behat/mink-extension": "~2.0",
 		"behat/mink-goutte-driver": "~1.1",
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
 		"libphutil": "1.2"
 	},
 	"require-dev": {
-		"behat/behat": "v3.1.0rc1",
+		"behat/behat": "~3.1@RC",
 		"behat/mink": "~1.6",
 		"behat/mink-extension": "~2.0",
 		"behat/mink-goutte-driver": "~1.1",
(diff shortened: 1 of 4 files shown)


jakobw:
Update: It looks like it works now!
Two problems appeared and there might be more that did not pop up yet.
- creating sprints failed when the user token wasn't set. fixed in db86e90 by making it a part of the login
- the `aSprintExistsForTheProject` created projects without slugs if it was run before some other feature which broke the URLs. fixed in 7fb580a

(Scrutinizer, you ok?)

wmde-manicki:
I like the fixes but still I am not sure if running tests in random order is what we want to do.
@jakobw are you able to restart Scrutinizer's run? I believe it failing is foobar but it happened several times recently, and it is a bit annoying

jakobw:
@wmde-manicki getting access denied for Scrutinizer unfortunately. Why don't we want to them randomized? It's easy enough to reproduce and will force us to write tests that can run independently from one another.

wmde-manicki:
Let's do it! I'll only have one possibly irrelevant comment to the patch

composer.json
```diff
@@ -28,7 +28,7 @@
 		"libphutil": "1.2"
 	},
 	"require-dev": {
-		"behat/behat": "~3.0",
+		"behat/behat": "v3.1.0rc1",
```
wmde-manicki: That is a bit harsh.
Wouldn't `~3.1` make composer install this 3.1.0 RC given https://github.com/Behat/Behat/blob/master/composer.json#L50 ? That is also not perfect as this would basically install master version, so I am not sure what is a better way.
So I am actually not opposing here.

jakobw: Agreed. I get
```
  Problem 1
    - Installation request for behat/behat ~3.1 -> satisfiable by behat/behat[v3.1.0rc1].
    - Removal request for behat/behat == 3.1.0.0-RC1 ```

when setting it to `~3.1`. `>=3.1.0rc1` works though. I'm not sure if I'm using composer wrong. Any ideas?

PR #203, opened 2016-02-05
Add tests to verify that most recent column change is considered when generating burndown chart

wmde-manicki:
Task: https://phabricator.wikimedia.org/T103071

As I understand the problem described in Phabricator ticket, at some point Phragile considered tasked being closed at the point it first got moved to the "closed" column, even if it has been moved back to "not closed" column afterwards.
If I am not mistaken this behaviour has been (possibly unintentionally) fixed along some other changes at some point (didn't manage to find the actual change).
`BurndownChartTest::testClosedPerDayConsidersMostRecentColumnChangeInWorkboardMode` is about checking the behaviour I described above. Also adding tests for two other classes involved that prove that the most recent column-change transaction is considered.

If I am wrong in what the Phab task is about, please comment!

### DIFF
--- a/tests/unit/StatusByWorkboardDispatcherTest.php
+++ b/tests/unit/StatusByWorkboardDispatcherTest.php
@@ -1,0 +1,159 @@
+<?php
+
+use Phragile\PhabricatorAPI;
+use Phragile\ProjectColumnRepository;
+use Phragile\StatusByWorkboardDispatcher;
+use Phragile\TransactionList;
+
+/**
+ * @covers Phragile\StatusByWorkboardDispatcher
+ */
+class StatusByWorkboardDispatcherTest extends TestCase {
+
+	private $workboardColumns = [
+		'PHID-123abc' => 'backlog',
+		'PHID-321cba' => 'done',
+		'PHID-abc123' => 'to do',
+	];
+
+	private function newPhabricatorAPI()
+	{
+		$phabricatorAPI = $this->getMockBuilder( PhabricatorAPI::class )
+			->disableOriginalConstructor()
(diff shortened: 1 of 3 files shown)


tests/unit/StatusByWorkboardDispatcherTest.php
```diff
@@ -0,0 +1,119 @@
...
+	{
+		$phabricatorAPI = $this->getMockBuilder( PhabricatorAPI::class )
+			->disableOriginalConstructor()
+			->getMock();
+		$phabricatorAPI->method('queryPHIDs')->will($this->returnCallback(function()
+		{
+			return array_map(function($column)
+			{
+				return ['name' => $column];
+			}, $this->workboardColumns);
```
WMDE-Fisch: you could also use array_flip( $this->columnPHIDs ) and get rid of the field

WMDE-Fisch:
+ 1,99

tests/unit/StatusByWorkboardDispatcherTest.php
wmde-manicki: True, I had that before like:
```php
return array_map(function($column)
{
	return ['name' => $column];
}, array_flip($this->workboardColumns));
```
but looking at this made me look like Jeroen's "double space OMG" cat.
But it doesn't really read that bad, so I'd probably change it the way you suggested.

wmde-manicki:
So I adjusted a bit `StatusByWorkboardDispatcher`'s tests to address @WMDE-Fisch's suggestion, plus split a length test method into three separate methods so tests stick better to Arrange-Act-Assert.

PR #207, opened 2016-02-05
Rename TransactionList to SortedTransactionList

wmde-manicki:
First I wanted to move sorting code from public method returning list of transactions. This has been against the CQS principle, actually.
Then I realized that `TransactionList` is only returning list of transactions sorted by time. So I renamed it (`TransactionListSortedByDate` or something similar might be even more accurate but it is getting too long) and its public method.
If there are more things coming to `TransactionList`, then renaming might be wrong. I am not aware of such plans, though.

This should be probably merged after https://github.com/wmde/phragile/pull/203 as that PR provides unit test for the class.


jakobw:
Makes sense!

PR #209, opened 2016-02-08
 Add tests for ProjectColumnRepository 

wmde-manicki:


tests/unit/ProjectColumnRepositoryTest.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+use Phragile\PhabricatorAPI;
+use Phragile\ProjectColumnRepository;
+
+/**
+ * @covers ProjectColumnRepositoryTest
```
addshore: covers tag is wrong! :(

PR #212, opened 2016-02-15
Discontinue certificate

jakobw:
Task: https://phabricator.wikimedia.org/T124032
* removed all Conduit certificate code
* globally catches a `ConduitException` with an error code of `ERR-INVALID-SESSION` and will tell the user that the Conduit API token is missing in the .env file.

I wanted to write a behat test for the Exception but couldn't think of a good way to dynamically override the `PHABRICATOR_API_TOKEN` environment variable in the test.


WMDE-Fisch:
+1,99 also cant think of a way to test this though.

PR #215, opened 2016-02-18
Maniphest search

jakobw:
Task: https://phabricator.wikimedia.org/T127179

Things this patch includes:
* use `maniphest.search` instead of `maniphest.query´
* introduce a `TaskDataFetcher` that replaces what were previously API calls to fetch tasks
* introduce a `TaskDataProcessor` (anyone got an idea for a better name?) that creates a list of `Task` objects from the output of `TaskDataFetcher`
* introduce `Task` which is a value object for tasks
* remove a bunch of stuff from `TaskList` and `TaskListTest` and put it where it belongs (mostly `TaskDataProcessor`, `StatusByWorkboardDispatcher` and their tests)
* test some more things
* `StatusByStatusFieldDispatcher` got a little bit uglier since information whether a task is closed is not included in `maniphest.search` so it now checks for a list of hardcoded statuses which might be silly

... whew, did I just rewrite Phragile?

These commits should probably be squashed before merging.

### DIFF
--- a/app/Phragile/PhabricatorAPI.php
+++ b/app/Phragile/PhabricatorAPI.php
@@ -47,14 +47,19 @@ class PhabricatorAPI {
 		)['data']);
 	}
 
-	public function queryTasksByProject($projectPHID)
+	/**
+	 * Uses maniphest.search to search tasks for a project.
+	 * @param $projectPHID
+	 * @return array
+	 */
+	public function searchTasksByProjectPHID($projectPHID)
 	{
 		return $this->client->callMethodSynchronous(
-			'maniphest.query',
+			'maniphest.search',
 			[
-				'projectPHIDs' => [$projectPHID]
+				'constraints' => ['projects' => [$projectPHID]]
 			]
-		);
+		)['data'];
 	}
 
 	public function createTask($projectPHID, array $task)
--- a/app/Phragile/PhabricatorAPI.php
+++ b/app/Phragile/PhabricatorAPI.php
@@ -49,17 +49,20 @@ class PhabricatorAPI {
 
 	/**
 	 * Uses maniphest.search to search tasks for a project.
-	 * @param $projectPHID
+	 * @param string $projectPHID
+	 * @param string $after - Cursor parameter to fetch the next page of tasks
 	 * @return array
 	 */
-	public function searchTasksByProjectPHID($projectPHID)
+	public function searchTasksByProjectPHID($projectPHID, $after)
 	{
 		return $this->client->callMethodSynchronous(
 			'maniphest.search',
 			[
-				'constraints' => ['projects' => [$projectPHID]]
+				'constraints' => ['projects' => [$projectPHID]],
+				'attachments' => ['projects' => true],
+				'after' => $after
 			]
-		)['data'];
+		);
 	}
 
 	public function createTask($projectPHID, array $task)
--- a/app/Phragile/Task.php
+++ b/app/Phragile/Task.php
@@ -1,0 +1,116 @@
+<?php
+namespace Phragile;
+
+class Task {
+	private $title;
+	private $priority;
+	private $points;
+	private $status;
+	private $closed;
+	private $id;
+	private $assigneePHID;
+	private $assigneeName;
+	private $cssClass;
+
+	function __construct($attributes)
+	{
+		$fields = ['title', 'priority', 'points', 'status', 'closed', 'id', 'assigneePHID'];
+		foreach ($fields as $field)
+		{
+			if (!key_exists($field, $attributes))
+			{
+				throw new \InvalidArgumentException("The $field field is missing.");
+			}
+
+			$this->$field = $attributes[$field];
+		}
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getTitle()
+	{
+		return $this->title;
+	}
+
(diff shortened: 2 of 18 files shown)


wmde-manicki:
I've just started to look how it works (on local instance) and seems it works fine for new sprints! Might be just my impression but fetching data from Phabricator seems to be slower (it takes some time before graphs are plotted etc). Have you also noticed something similar?

When selecting sprint that has ended (thus snapshot data is being fetch) I am getting the following error message: `ErrorException in TaskDataProcessor.php line 24: Undefined index: fields`. Or at least my guess it is related to snapshots. I believe this is expected and it is going to be fixed in https://phabricator.wikimedia.org/T127180, right?

I haven't really looked at the code yet.

jakobw:
Yes, it seemed slower for projects with >100 tasks which I guess makes sense since it now does multiple HTTP requests for those.

Correct, snapshots aren't supposed to work yet since they don't have the JSON structure that the `TaskDataProcessor` is expecting.

app/Phragile/PhabricatorAPI.php
```diff
@@ -47,12 +47,20 @@ private function queryProject($fields)
 		)['data']);
 	}
 
-	public function queryTasksByProject($projectPHID)
+	/**
+	 * Uses maniphest.search to search tasks for a project.
+	 * @param $projectPHID
```
wmde-manicki: it is a string, right?

jakobw: Yes! Fixed

app/Phragile/Task.php
```diff
@@ -0,0 +1,116 @@
...
+	private $assigneePHID;
+	private $assigneeName;
+	private $cssClass;
+
+	function __construct($attributes)
+	{
+		$fields = ['title', 'priority', 'points', 'status', 'closed', 'id', 'assigneePHID'];
+		foreach ($fields as $field)
+		{
+			if (!key_exists($field, $attributes))
```
wmde-manicki: in modern PHP versions this function is called `array_key_exists` :)

app/Phragile/Task.php
```diff
@@ -0,0 +1,116 @@
...
+	private $priority;
+	private $points;
+	private $status;
+	private $closed;
+	private $id;
+	private $assigneePHID;
+	private $assigneeName;
+	private $cssClass;
+
+	function __construct($attributes)
```
wmde-manicki: If I didn't miss any place where the instance of this class is being created, these `$attributes` are always an array. We could have a type hint here then.

app/Phragile/Task.php
```diff
@@ -0,0 +1,116 @@
...
+	{
+		$fields = ['title', 'priority', 'points', 'status', 'closed', 'id', 'assigneePHID'];
+		foreach ($fields as $field)
+		{
+			if (!key_exists($field, $attributes))
+			{
+				throw new \InvalidArgumentException("The $field field is missing.");
+			}
+
+			$this->$field = $attributes[$field];
```
wmde-manicki: this could make it difficult to find where exactly value of `$this->title` or `$this->closed` is set. But as long as the class and its constructor are that compact, I wouldn't oppose to have it like this here.

app/Phragile/TaskDataFetcher.php
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	/**
+	 * @param string $projectPHID
+	 * @return array
+	 */
+	public function fetchProjectTasks($projectPHID)
+	{
+		$rawTaskData = [];
+		$after = 0;
```
wmde-manicki: to be super-accurate this could be '0' (a string, not a number).

app/Phragile/TaskDataProcessor.php
```diff
@@ -0,0 +1,50 @@
...
+	{
+		$this->statusDispatcher = $statusDispatcher;
+		$this->options = $options;
+	}
+
+	/**
+	 * @param $rawData
+	 * @return Task[]
+	 */
+	public function process($rawData)
```
wmde-manicki: `array $rawData` + type in doc above

wmde-manicki:
So I had a look at the code and:
 - I really like `Task` class being introduced, and moving out to separate classes stuff that is doing the magic related to processing data from Phabricator,
 - found couple of small things that @jakobw already managed to fix before I even finished reviewing the patch!
 - I agree that having "closed statuses" hard-coded in `StatusByStatusFieldDispatcher` is no good but @jakobw has already idea how to make it clean and generic (see https://phabricator.wikimedia.org/T127433). I am not sure if this must be included in this patch, or added later. I'd be fine if it is changed later,
 - having looked at the patch I started to think that maybe some parts of the code dealing with Phabricator could be moved out from the `Sprint` class. But this is definitely beyond the scope here.

So basically I'd consider this PR as ready to merge if @jakobw didn't mention cleaning up the commit list a bit. At least "Fix foo" commits could be probably joined with relevant commits. Then I'd be happy to merge. Very nice job!

jakobw:
@wmde-manicki Squashed the commits!

I also think it's fine to fix the status issue later and I agree that some of the stuff in the `Sprint` class should probably go elsewhere, especially since it's pretty redundant with some of the code in `SprintDataFactory`.

PR #216, opened 2016-02-19
Maniphest.query -> maniphest.search migration script

jakobw:
Task: https://phabricator.wikimedia.org/T127180

Can be executed via `php artisan snapshots:migrate`.
The tests look a bit ugly since those JSON strings/objects are so awkward to move around.

### DIFF
--- a/app/Console/Commands/MigrateSnapshots.php
+++ b/app/Console/Commands/MigrateSnapshots.php
@@ -1,0 +1,30 @@
+<?php namespace App\Console\Commands;
+
+use App\Console\Commands\Lib\SnapshotDataConverter;
+use Illuminate\Console\Command;
+use SprintSnapshot;
+
+class MigrateSnapshots extends Command {
+
+	protected $name = 'snapshots:migrate';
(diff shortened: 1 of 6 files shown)


wmde-manicki:
Looks fine, and it also converted my old snapshot so they don't cause Phragile to crash.

My only (minor) concern is the name of the new class. `SnapshotDataConverter` and its public method's name `convert` don't really tell what kind of conversion is in question. For instance class name could give more specific information what is converted. On the other hand, as this is the only conversion in Phragile's history so far, it is rather clear what's going on.

That said, I couldn't come up with some other name to suggest that wouldn't be more than thirty-characters-long, so I think current names are also fine! I am just thinking out loud.

I'd add a commit fixing indentation in behat feature file. Apart from that PR is good.

@WMDE-Fisch @addshore willing to have a look on this too?

jakobw:
Fixed the indentation issue!
I'm open to suggestions for a better class name! :)
If we can't come up with anything better that doesn't read like a sentence (`ManiphestQuerySnapshotToManiphestSearchFormatConverter`? :D) we could can add a comment explaining what the class is for.

manicki:
Looking good now, thanks!
And I think comment will be fine.

PR #217, opened 2016-03-02
fix story point test

WMDE-Fisch:
The old test passed although story points could not be retrieved due to the fact, that the tested text for the points was found elsewhere on the page. ( e.g. like in `#13` before the task title )

This PR fixes that by introducing a new test and adjusting the story point settings in the test environment as well as in the default config file.

https://phabricator.wikimedia.org/T128726

### DIFF
--- a/tests/acceptance/bootstrap/FeatureContext.php
+++ b/tests/acceptance/bootstrap/FeatureContext.php
@@ -224,17 +224,18 @@ class FeatureContext extends MinkContext implements Context, SnippetAcceptingCon
 	}
 
 	/**
-	 * @Given it has the following tasks for the :sprint Sprint:
+	 * @Given :sprint has a task with :title :priority and :points
 	 */
-	public function itHasTheFollowingTasks($sprint, \Behat\Gherkin\Node\TableNode $table)
+	public function sprintHasATaskWith($sprint, $title, $priority, $points)
 	{
 		$sprintPHID = Sprint::where('title', $sprint)->first()->phid;
 		$phabricator = App::make('phabricator');
 
-		foreach ($table->getHash() as $task)
-		{
-			$phabricator->createTask($sprintPHID, $task);
-		}
+		$this->selectedTask = $phabricator->createTask($sprintPHID, [
+			'title' => $title,
+			'priority' => $priority,
+			'points' => $points
+		]);
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


manicki:
I agree the current test is wrong. And I generally like changes made here but: now we would have two tests "See tasks on the Sprint Backlog" and the new "See task details in Sprint Backlog" which are basically about the same things.
How about adjusting what you do here slightly:
1. Do not add new test
2. Use "I should see the task with :points story points", "I should see the task with :title as title" in the exisiting tests (probably also add similar thing for priority) but don't base on the task id there, ie. use something like `#backlog .points` to `#tID .points` get points, etc?

How does it sound like? Will that work?
Probably remembering last added test is not need as well then.

manicki:
Oh no, forget that comment for now. This css selector has to only match a single element, so that wouldn't work. Maybe something like passing a task number ("first", "second" task) would help, then `nth-child` could be maybe used. I'll think about it more than few minutes and get back here.

WMDE-Fisch:
changed the test to Scenario Outline and check different tasks added individually

tests/acceptance/bootstrap/FeatureContext.php
```diff
@@ -224,17 +224,18 @@ public function aSprintExistsForTheProject($sprintTitle, $projectTitle)
 	}
 
 	/**
-	 * @Given it has the following tasks for the :sprint Sprint:
+	 * @Given :sprint has a task with :title :priority and :points
 	 */
-	public function itHasTheFollowingTasks($sprint, \Behat\Gherkin\Node\TableNode $table)
+	public function itHasTheFollowingTasks($sprint, $title, $priority, $points)
```
manicki: method could be possible renamed but it a minor issue, I think

tests/acceptance/bootstrap/FeatureContext.php
```diff
@@ -509,6 +510,30 @@ public function iShouldSeeMyNameInTheTaskSRowOfTheSprintBacklog()
...
+	{
+		$this->assertElementContains('#t' . $this->selectedTask['id'] . ' .title', $title);
+	}
+
+	/**
+	 * @Then I should see the task with :priority as priority
+	 */
+	public function iShouldSeeTheTaskWithPriority($priority)
+	{
+		$this->assertElementContains('#t' . $this->selectedTask['id'] . ' .filter-backlog', $priority);
```
manicki: didn't you add `priority` class to the relevant `td` element to get priority here? This then should rather use `.priority` instead of `.filter-backlog`

WMDE-Fisch:
damn - yeah fixed both issues -.-

manicki:
Apart from CSS class issue that has already been address, this looks good!

I'd suggest some clean up of this pull request, though:
 - as this is no longer adding two different tests but rather making the existing one more accurate it seems to me that https://github.com/wmde/phragile/commit/e285cd6d4d7701f83dfd87b8f3999e26a96267f0, https://github.com/wmde/phragile/commit/5dffa3243d3388cceab6b128d430a8b1f149ca12, https://github.com/wmde/phragile/commit/554439577081287b2ef858a6de0f5b912746943d, https://github.com/wmde/phragile/commit/bf560388d49b4bc2d5b80dd55ee41c375d2ead9c, https://github.com/wmde/phragile/commit/c0fa0805c1012b886a63b2bec416163368102939 could be squashed into a single commit introducing scenario outline and changing the existing context method slightly, etc.
 - https://github.com/wmde/phragile/commit/5dffa3243d3388cceab6b128d430a8b1f149ca12 has no use as far as I can tell,
 - shouldn't changes from https://github.com/wmde/phragile/commit/9afb6ab38b02fb303ed6a9f016c6e710db569ac3 and https://github.com/wmde/phragile/commit/3292e546ca4e6196a57be37d6058313b9cf4895f be actually introduced in the single commit? It seems to me that having them separate would lead to the situation where some things are broken on one point, and some other things on the other
 - minor but as talking rebasing here: https://github.com/wmde/phragile/commit/2317edca6a15b506a66c04a65be5b14c576a46f3 could be possibly also squashed into https://github.com/wmde/phragile/commit/795d04744c2dabdd61693b0aadce34126ab09346

WMDE-Fisch:
clean it all :->!

manicki:
nice!

PR #218, opened 2016-03-03
Show assignees in sprint backlog table

manicki:
Task: https://phabricator.wikimedia.org/T128717

Along with https://github.com/wmde/phragile/pull/215 fetching assignees got kind of broken. This does not become obvious when there are only 100 or fewer Phabricator users, that's why it hasn't been noticed by already existing Behat test.

Namely `SprintDataFactory` has been always passing `AssigneeRepository` an empty array, which the repository in turn used in Phabricator API query. When given empty list of user PHIDs API returns all users but truncates the list to first 100. In the test environment there are only few users, so the repository always got needed user but in the production environment this bug results with showing no assignees, see below (at least some of those tasks has already an user assigned):
phragile-no-assignees

### DIFF
--- a/app/Phragile/AssigneeRepository.php
+++ b/app/Phragile/AssigneeRepository.php
@@ -11,7 +11,7 @@ class AssigneeRepository {
 
 	/**
 	 * @param PhabricatorAPI $phabricator
-	 * @param array[] $tasks
+	 * @param Task[] $tasks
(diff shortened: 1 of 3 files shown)


tobijat:
Yay for adding the test!
And Yay for fixing the problem!
thx!

PR #221, opened 2016-03-07
improved snapshot script to handle big snapshot data

WMDE-Fisch:
Migration on Labs failed due to the number of snapshots there (~2300). Loading all the snapshots with `SprintSnapshot::all()` consumed to much memory and the script failed.

see https://phabricator.wikimedia.org/T128581

The script was refactored to read the snapshots in batches and migrate them one after the other. In the process additional command line options were added e.g. to set step size. Simple output gives feedback on the progress of the migration.

### DIFF
--- a/app/Console/Commands/MigrateSnapshots.php
+++ b/app/Console/Commands/MigrateSnapshots.php
@@ -2,6 +2,7 @@
 
 use App\Console\Commands\Lib\SnapshotDataConverter;
 use Illuminate\Console\Command;
+use Symfony\Component\Console\Input\InputOption;
 use SprintSnapshot;
 
 class MigrateSnapshots extends Command {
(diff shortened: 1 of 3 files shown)


manicki:
Apart from some minor things (spacing and output) looks good.

I am not sure if we couldn't come up with more precise variable name than "stepSize" but at least I would rephrase the confirmation message (as in the relevant comment).
I am also not convinced that the "force" option is really necessary. Migration would be only run along with Phragile/Phabricator update, and the migration would only be run by person having an access to the Phragile instance (ie. it is rather safe to assume by person who knows what she's doing).But this is just old man's ranting, that's not a big deal, and it is probably better to have more confirmation requests than fewer.

app/Console/Commands/MigrateSnapshots.php
```diff
@@ -11,20 +12,63 @@ class MigrateSnapshots extends Command {
...
+	/**
+	 * Get the console command options.
+	 *
+	 * @return array
+	 */
+	protected function getOptions()
+	{
+		return [
+			['force', null, InputOption::VALUE_NONE, 'Force the operation to run when in production.'],
+
```
manicki: the blank line could be possibly omitted

WMDE-Fisch:
rebased, fixed code styles
renamed to batchSize, makes more sense in the context I guess
`--force` option is mainly used for tests to skip confirmation

WMDE-Fisch:
next try -.-

PR #222, opened 2016-03-08
updated phpcs rules and fixed code

WMDE-Fisch:
ignored inline control structures and namspace check
removed unused methods / classes
removed function calls in loops
fixed styling
add cs checks to ci


composer.json
```diff
@@ -69,7 +69,8 @@
 		],
 		"ci": [
 			"composer validate --no-interaction",
-			"vendor/bin/phpunit"
+			"vendor/bin/phpunit",
+			"@phpcs"
```
manicki: is this `@thing` some kind of alias to `composer thing`?

manicki: ah, no, forget the question. I see this the reason of CI build's failure. So this line should be changed to `composer phpcs`, right?

tests/acceptance/bootstrap/FeatureContext.php
```diff
@@ -77,11 +77,13 @@ private function phabricatorLogin()
 		} catch(Behat\Mink\Exception\ElementNotFoundException $e)
 		{
 			// Absence of this button just means that the user has authorized Phragile before.
+			return;
```
manicki: I find it odd to have stuff in `finally` block but not execute it. But after looking at this method for some minutes it looks like this should be addressed separately.

manicki:
So it looks like updating of composer.lock is not definitely required. But I'd let it be here, it shouldn't do any harm.

PR #225, opened 2016-03-08
composer package updates

WMDE-Fisch:
Checked the status of the libs used. `illuminate/html` is deprecated and got replaced by `laravelcollective/html`. The version should meet the laravel version used.

Since Laravel 5.1 needs PHP >=5.5.9 and we have Travis checking >=5.5 this should have been updated back then already.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
 	"require": {
 		"php": ">=5.4.0",
 		"laravel/framework": "5.1.*",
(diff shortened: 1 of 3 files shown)


manicki:
+1. I have no idea about those `Illuminate\HTML` and `Collective\HTML` things but it looks sane!

manicki:
Needs a rebase

PR #226, opened 2016-03-09
Adjust PHPCS rules

manicki:
Following up https://github.com/wmde/phragile/pull/222 some more PHPCS adjustments to make code reviews easier.
I don't think it would be valid to say that we're aiming to stick to Laravel Code Style as they have changed their guidelines to be more PSR-2-alike. Any way rules added here are intended to make code follow the code style that Phragile uses from its beginning. Let's have the Phragile Code Style!

I've re-enabled sniff complaining about evil in-line control structures, and decreased a line length limit to 125 as used in many other WMDE repos.
https://github.com/wmde/phragile/commit/dfd01937cb104e302e45b20c5413bdc9cdaeddac takes care of not allowing `if ( sth )`.
https://github.com/wmde/phragile/commit/3acb64eca27d1ea20d0f6a3976e914366fab3dd1 is for things like
```php
foreach ($foo)
{
foo()
}
```
and
```php
if ($cat === 'nyan')
{
foo();
} else
{
bar();
}
```

Still there are some things that are missing, especially PHPCS wouldn't complain when function calls are written in the MW-alike spacey way e.g. `foo( 'bar', 'baz', 666 )` instead of preferred `foo('bar', 'baz', 666)`. I'd also like to have PHPCS enforce spaces around concatenation operator but it is missing at the moment. I'd try to add checks for those later.

### DIFF
--- a/app/Console/Commands/MigrateSnapshots.php
+++ b/app/Console/Commands/MigrateSnapshots.php
@@ -19,7 +19,7 @@ class MigrateSnapshots extends Command {
 			$confirmation = $this->confirm(
 				'Do you really want to migrate ' . $snapshotCount . ' snapshots in batches of ' . $batchSize . '?'
 			);
-			if ( !$confirmation) {
+			if (!$confirmation) {
 				$this->line('Migration aborted.');
 				return;
 			}
--- a/app/Console/Commands/MigrateSnapshots.php
+++ b/app/Console/Commands/MigrateSnapshots.php
@@ -16,7 +16,9 @@ class MigrateSnapshots extends Command {
 		$batchSize = intval($this->input->getOption('batchSize'));
 
 		if (!$this->input->getOption('force')) {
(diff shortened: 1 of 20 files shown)


manicki:
Apparently I have messed something up as CI builds fail. I'll have a look into that.

manicki:
I've rebased the original PR. As far as I can tell, behat test failure is not related to changes done here. It looks like there might be a problem with loading a snapshot data in some cases. Possible fix submitted in: https://github.com/wmde/phragile/pull/229. After fixing the issue with tests this PR's branch will be rebased again!

manicki:
It seems tests still fail. I am out of ideas what is causing this for now. I'll get back to that in a bit.

manicki:
I've finally managed to find a reason of two behat tests failing in the previous version of the patch. It has indeed been my fault not something different. Silly me!

WMDE-Fisch:
Yay! Looks good and thanks for digging into this!

PR #235, opened 2016-03-29
Update release notes.

jakobw:
Release 2.0.0!

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,10 @@
 # Phragile release notes
 
+## Version 2.0.0 (2016-03-29)
+* Support Phabricator Projects v3
+* Use maniphest.search instead of maniphest.query
+* Provide migration script for snapshot conversion
+
 ## Version 1.1 (2016-02-15)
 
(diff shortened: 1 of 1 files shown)


tobijat:
+1

RELEASE-NOTES.md
```diff
@@ -1,5 +1,10 @@
 # Phragile release notes
 
+## Version 2.0.0 (2016-03-29)
+* Support Phabricator Projects v3
+* Use maniphest.search instead of maniphest.query
+* Provide migration script for snapshot conversion
```
JeroenDeDauw: I think it's wrong to use imperative here. (Same goes for commits, though there it seems more debatable.) The audience of release notes is users that want to know what was changed. Hence past tense seems appropriate, ie `Added support for Phabricator Projects v3` or `Provided migration script for snapshot conversion`.

Also keep in mind that changes to internal workings not relevant to such users just confuse them. The second point here might fall into that category (just guessing here, I don't know what it means).

WMDE-Fisch: +1

PR #236, opened 2016-04-08
Fix priority filtering.

jakobw:
Task: https://phabricator.wikimedia.org/T132143


resources/js/sprint_backlog.js
```diff
@@ -19,7 +19,9 @@
             var values = item.values();
 
             for (var c in currentFilters) {
-                if (values[c].trim() != currentFilters[c]) {
+                if (c === 'priority' && $('<div/>').append(values[c]).find('span').text().trim() != currentFilters[c]) {
+                    return false;
+                } else if (c !== 'priority' && values[c].trim() != currentFilters[c]) {
```
JeroenDeDauw: How about extracting the predicate(s)?

PR #237, opened 2016-04-18
updated release notes for Phabricator dependency

WMDE-Fisch:
set to the version used at Wikimedia when switching to Projects v3 
see https://phabricator.wikimedia.org/T120013#2037098

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,14 @@ For more information see the [product backlog](https://github.com/wmde/phragile/
 
 * PHP 5.5 or later
 * MySQL, SQLite or PostgreSQL
+* Phabricator (see below)
 
 ### Preparation
 
 * [Install Phabricator](https://secure.phabricator.com/book/phabricator/article/installation_guide/)
+    
+    Phragile >=2.0.0 requires Phabricator release from 2016 Week 8 or newer
+
 * [Activate Phabricator OAuth](https://github.com/wmde/phragile/wiki/Activating-Phabricator-OAuth)
 * Add a  custom field for story points to http://yourphabricator/config/edit/maniphest.custom-field-definitions/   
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,10 +30,14 @@ For more information see the [product backlog](https://github.com/wmde/phragile/
 
 * PHP 5.5 or later
 * MySQL, SQLite or PostgreSQL
+* Phabricator ( see below )
```
manicki: why the mediawiki spacing here?

WMDE-Fisch: hmm I guess I got used to it to much ^^

README.md
```diff
@@ -30,10 +30,14 @@ For more information see the [product backlog](https://github.com/wmde/phragile/
...
 
 ### Preparation
 
 * [Install Phabricator](https://secure.phabricator.com/book/phabricator/article/installation_guide/)
+    
+    Phragile >=2.0.0 needs Phabricator >=2016 Week 8
```
manicki: This is probably fine to have it written like that.
I would actually more verbose and put it like: "Phragile 2.0.0 requires Phabricator release from 2016 Week 8 or newer".
Current version of the text might become invalid quite easily, e.g. version 2.1.0 might require Phab release from Week 23, and version 3.0.0 Phab release from Week 666.
But yeah, I agree, this is a totally minor thing which will be change when needed. I am just babbling here to have a break of messing with MW core code :)

Offtopic: omg, they are using week numbers. Is it only me who has never any idea what it the week number of the current (or any other) week and has to always look it up?

WMDE-Fisch: I would see the readme should always say something about the latest version requirements and the rest will go and remain in the release notes

manicki: yea, you're right, I agree with this comment

PR #238, opened 2016-04-28
Make Phragile work with the new transaction format for columns.

jakobw:
Task: https://phabricator.wikimedia.org/T133855

This will fix sprint live views as well as new snapshots on Phragile + respective unit tests. However, old snapshots will be broken.

### DIFF
--- a/tests/unit/BurndownChartTest.php
+++ b/tests/unit/BurndownChartTest.php
@@ -181,7 +181,6 @@ class BurndownChartTest extends TestCase {
 			[
 				'1' => [[
 					'transactionType' => 'core:columns',
-					'oldValue' => null,
 					'newValue' => [[
 						'fromColumnPHIDs' => ['anyNotClosed' => 'anyNotClosed'],
 						'columnPHID' => $this->closedColumnPHIDs[1],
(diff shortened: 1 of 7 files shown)


tests/unit/BurndownChartTest.php
```diff
@@ -180,25 +180,23 @@ public function testClosedPerDayIgnoresStatusChangeInWorkboardMode()
...
-						'columnPHIDs' => ['anyNotClosed'],
-						'projectPHID' => $this->testProjectPHID,
-					],
-					'newValue' => [
-						'columnPHIDs' => [$this->closedColumnPHIDs[1]],
-					],
+					'transactionType' => 'core:columns',
+					'oldValue' => null,
+					'newValue' => [[
+						'fromPHIDs' => ['anyNotClosed' => 'anyNotClosed'],
```
manicki: shouldn't this key be `fromColumnPHIDs`?

jakobw: Yes! The one that's a couple of lines below should be as well. Will fix

manicki:
Checked the new version locally and works good! Also code changes look sane.
It looks like the `oldValue` field in the column-change transaction data is no longer used (always set to `null` apparently?). In most test cases it has been skipped (I think all changed tests except first two `BurndownChartTest`s) but I'd rather keep this `oldValue => null` for consistency with production data. I'll add a tiny commit adding missing empty fields and then merge this.

Let's see what explodes after it's merged!

manicki:
Nah, I've changed my mind. Test data does not mock several other transaction data fields that are irrelevant for Phragile. So I'd rather remove `oldValue` from those test cases where it is still present.

PR #239, opened 2016-05-02
[DNM]Fix transaction types in tests

manicki:
Those should have been also updated in https://github.com/wmde/phragile/pull/238 (for sake of consistency) but we missed them as transaction type is not really checked/validated/relevant in those tests.


manicki:
Nah, this is wrong. It is not only about changing the `transactionType`. Also `oldValue` becomes reduntant etc. Will fix this and remove the [DNM] prefix when ready!

manicki:
This seems no longer relevant given the changes that are about to be introduced in https://github.com/wmde/phragile/pull/240

PR #240, opened 2016-05-03
Use simplified transaction data structure internally and in snapshots

manicki:
It took me far more time than I expected but I've broken all things couple of times while I was changing exisiting stuff, but it is finally ready for others to have a look.

This PR is far from complete. There are multiple TODOs already marked in the code, and lots of others missing. I am submitting this a bit pre-maturely as a pull request as I hope to get opinions from others as soon as possible on the approach I have taken here.
So I am pinging @jakobw @addshore @WMDE-Fisch and all the people and ask for merciless review!

Idea: Currently transaction data is fetched from Phabricator API response and used both by Phragile classes as an internal data structure, and also used as a part of snapshots in unchanged form.
There is some filtering being done and only transactions relevant to status and column changes are taken over from Phabricator. Still, those data contain quite some information which is not used by Phragile. Also, whenever Phabricator changes its API there is a risk that the change will be a breaking change for Phabricator. This happened last week after changing the structure of column-change transactions in the API response (plus similar breaking change earlier this year, although related to tasks)

Approach I have used is basically stealing ideas from @jakobw's (task-related patch)[https://github.com/wmde/phragile/pull/215]. I have introduced "Processor" class that gets a json-decoded array from Phabricator and transforms the data into array of "*Transaction" class objects representing particular transaction types relevant to Phragile.
I think I went a bit further than @jakobw as here only classes that know anything of how looks API response are the Processor itself and the previously existing class filtering unwanted transactions (I think for example for performance reason it makes no sense to do the conversion before filtering). Also, arrays 

What is obviously missing here is a script to convert transaction data of already existing snapshots to the format which would be compatible with changes made here. It has to be done (and deployed) along with other changes. Given the already existing script converting task data, I would consider it a rather small remaining task.
Apart from that all normal operations of Phragile are supposed to be "generally" working:
 - live view of the sprint,
 - creating a new snapshot of the sprint using live data,
 - viewing the snapshot created _after_ this code has been introduced.

I have added some TODOs already. Some should be added to make the patch complete (e.g. testing of special cases), some will/would rather be considered in separate PRs.

This is a rather big patch, touching some like half of Phragile classes (probably exaggerated estimate), and I'd expect it to grow even bigger while missing things are added. I think it makes sense to do the general review of how things are being changed here first, focusing on the approach taken and discovering some obvious and critical flaws. Detailed whitespace-level review might be done once this is all finished functionality-wise.

Current division of the code introduced here into three commits is kind of artificial, I'd say as it all depends on each other. To make reviewing of the final change easier I'd try to divide it somehow, although it might be tricky.

I'd be glad to hear any comments on this rather big thing!

### DIFF
--- a/app/Phragile/ColumnChangeTransaction.php
+++ b/app/Phragile/ColumnChangeTransaction.php
@@ -4,6 +4,8 @@ namespace Phragile;
 
 class ColumnChangeTransaction extends Transaction {
 
+	const TYPE = 'columnChange';
+
 	/**
 	 * @var string
 	 */
@@ -60,8 +62,7 @@ class ColumnChangeTransaction extends Transaction {
 	public function getTransactionData()
 	{
 		return [
-			'type' => 'columnChange', // TODO: this string should be moved to some constant,
-			//                        //so it there is no need to type it in in other places. But constant of what? Transaction::sth?
+			'type' => self::TYPE,
 			'timestamp' => $this->timestamp,
 			'workboardPHID' => $this->workboardPHID,
 			'oldColumnPHID' => $this->oldColumnPHID,
--- a/app/Phragile/ColumnChangeTransaction.php
+++ b/app/Phragile/ColumnChangeTransaction.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Phragile;
+
+class ColumnChangeTransaction extends Transaction {
+
+	/**
+	 * @var string
+	 */
+	private $workboardPHID;
+
+	/**
+	 * @var string
+	 */
+	private $oldColumnPHID;
+
+	/**
+	 * @var string
+	 */
+	private $newColumnPHID;
+
+	// TODO: use array of values instead of n positional params?
+	/**
+	 * @param string $timestamp
+	 * @param string $workboardPHID
+	 * @param string|false $oldColumnPHID
+	 * @param string $newColumnPHID
+	 */
+	public function __construct($timestamp, $workboardPHID, $oldColumnPHID, $newColumnPHID)
+	{
+		$this->timestamp = $timestamp;
+		$this->workboardPHID = $workboardPHID;
+		$this->oldColumnPHID = $oldColumnPHID;
+		$this->newColumnPHID = $newColumnPHID;
+	}
+
+	public function getWorkboardPHID()
+	{
+		return $this->workboardPHID;
+	}
+
+	public function getOldColumnPHID()
+	{
+		return $this->oldColumnPHID;
+	}
+
+	public function getNewColumnPHID()
+	{
+		return $this->newColumnPHID;
+	}
+
+	public function getTransactionData()
+	{
+		return [
+			'type' => 'columnChange', // TODO: this string should be moved to some constant,
+			//                        //so it there is no need to type it in in other places. But constant of what? Transaction::sth?
+			'timestamp' => $this->timestamp,
+			'workboardPHID' => $this->workboardPHID,
+			'oldColumnPHID' => $this->oldColumnPHID,
+			'newColumnPHID' => $this->newColumnPHID,
+		];
+	}
+
+}
--- a/app/Phragile/ColumnChangeTransaction.php
+++ b/app/Phragile/ColumnChangeTransaction.php
@@ -19,19 +19,27 @@ class ColumnChangeTransaction extends Transaction {
 	 */
 	private $newColumnPHID;
 
-	// TODO: use array of values instead of n positional params?
 	/**
-	 * @param string $timestamp
-	 * @param string $workboardPHID
-	 * @param string|false $oldColumnPHID
-	 * @param string $newColumnPHID
+	 * @param array $attributes associative array with keys:
+	 *        - 'timestamp' string
+	 *        - 'workboardPHID' string
+	 *        - 'oldColumnPHID' string|false false means no old column (initial setting of the column)
+	 *        - 'newColumnPHID' string
 	 */
-	public function __construct($timestamp, $workboardPHID, $oldColumnPHID, $newColumnPHID)
+	public function __construct(array $attributes)
 	{
-		$this->timestamp = $timestamp;
-		$this->workboardPHID = $workboardPHID;
-		$this->oldColumnPHID = $oldColumnPHID;
-		$this->newColumnPHID = $newColumnPHID;
+		$fields = ['timestamp', 'workboardPHID', 'oldColumnPHID', 'newColumnPHID'];
+		foreach ($fields as $field)
+		{
+			if (!array_key_exists($field, $attributes))
+			{
+				throw new \InvalidArgumentException('The ' . $field . ' field is missing.');
+			}
+		}
+		$this->timestamp = $attributes['timestamp'];
+		$this->workboardPHID = $attributes['workboardPHID'];
+		$this->oldColumnPHID = $attributes['oldColumnPHID'];
+		$this->newColumnPHID = $attributes['newColumnPHID'];
 	}
 
 	public function getWorkboardPHID()
--- a/app/Phragile/Transaction.php
+++ b/app/Phragile/Transaction.php
@@ -4,6 +4,8 @@ namespace Phragile;
 
 abstract class Transaction {
 
+	const TYPE = 'undefined';
+
 	// TODO: consider renaming the field to something more meaningful
 	protected $timestamp;
 
--- a/app/Phragile/Transaction.php
+++ b/app/Phragile/Transaction.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Phragile;
+
+abstract class Transaction {
+
+	// TODO: consider renaming the field to something more meaningful
+	protected $timestamp;
+
+	public function getTimestamp()
+	{
+		return $this->timestamp;
+	}
+
+	abstract public function getTransactionData();
+
+}
--- a/app/Phragile/Transaction.php
+++ b/app/Phragile/Transaction.php
@@ -12,6 +12,7 @@ abstract class Transaction {
 		return $this->timestamp;
 	}
 
+	// TODO: would getSnapshotData be a better name?
 	abstract public function getTransactionData();
 
 }
--- a/app/Phragile/TransactionFilter.php
+++ b/app/Phragile/TransactionFilter.php
@@ -24,12 +24,12 @@ class TransactionFilter {
 
 	protected function isWorkboardTransaction(array $transaction)
 	{
-		return $transaction['transactionType'] === 'core:columns';
+		return $transaction['transactionType'] === TransactionRawDataProcessor::COLUMN_CHANGE_TRANSACTION;
 	}
 
 	protected function isStatusTransaction(array $transaction)
 	{
-		return $transaction['transactionType'] === 'status'
-		    || $transaction['transactionType'] === 'mergedinto';
+		return $transaction['transactionType'] === TransactionRawDataProcessor::STATUS_CHANGE_TRANSACTION
+		    || $transaction['transactionType'] === TransactionRawDataProcessor::MERGED_INTO_TRANSACTION;
 	}
 }
--- a/app/Phragile/TransactionFilter.php
+++ b/app/Phragile/TransactionFilter.php
@@ -1,6 +1,7 @@
 <?php
 namespace Phragile;
 
+// TODO: rename to e.g. TransactionDataFilter
 class TransactionFilter {
 	/**
 	 * Filters out irrelevant transactions
--- a/app/Phragile/TransactionLoader.php
+++ b/app/Phragile/TransactionLoader.php
@@ -1,6 +1,7 @@
 <?php
 namespace Phragile;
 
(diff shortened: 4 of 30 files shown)


manicki:
And as final words: given that we decide to move from Phabricator-based data structure to some Phragile own ones (be it one suggested here or any other), I started to think what should we do to avoid problems when Phabricator API changes unexpectedly for us, as it has already happen. Currently when there is a breaking change in Wikimedia Phabricator, we noticed because either live data looks suspicious/totally wrong and/or snapshot data is incomplete/broken. Each time we first did a change to the "live" version and then fixed previously broken snapshots in some way later.
I'd consider "broken" snapshots a good indicator that something changed in the API and Phragile does no longer get all the data it expects. What might be worth considering is having some kind of a light-weight tool that checks, once snapshot creating cron job has run, that there are no suspicious things in snapshots. That might at least sound promising but I am not really sure what those "suspicious things" could be to have a reliable results (no false alarms every other day), so the tool would really be light-weight.
Possible indicators could be for example:
 - empty task data in all snapshots,
 - no transaction data for any task in all snapshots.
This would only prevent from total disaster but maybe this is what we want?

Also, I was thinking is it better in case of breaking API change to have some broken snapshots that can be (possibly) fixed later, or have no snapshots in the "broken" period at all. Assuming the time period needed to update Phragile is rather short, the latter might be acceptable solution (at least from my perspective, I am aware that for some users having all the snapshots might be crucial). I am mentioning this also because once we decide not to store data received from API there might be 
no way to fix broken snapshots when there is breaking change in the API. Ie. when the conversion from API structure to Phragile's internal structure fails it will just put empty data to the snapshot and I don't see a way to restore it later. I think this should be also taken into consideration when deciding which way we want to got with transaction data in snapshots etc.

This comment doesn't real add much of the value here, but I am just leaving it here as a note before I forgot all of this.

WMDE-Fisch:
I generally like the idea of have or own data structure and the stuff done in this PR. 

Related to the question of "spontaneous" upgrades of Phabricator: This is / should not be a real problem. Phabricator upgrades should be well prepared and we, as the maintainers of Phragile should be included in the planing process or at least be notified. Then we can check and adjust before anything breaks or gets lost. 

I agree, we should definitely develop a method to detect incompatible API responses and show an error to the user &/ disable certain options. For non-Wikimedia users we could also add a hint in the readme, that Phabricator updates can be incompatible to the current Phragile version and should be tested before used.

manicki:
> I agree, we should definitely develop a method to detect incompatible API responses and show an error to the user &/ disable certain options.

Actually, I wouldn't add this. I think currently Phragile behaviour is quite ok when those API breaking changes happen. It just shows no/partial data. That indeed makes Phragile graphs look strange but I think it's fine. It clearly indicates something is wrong.
I would not invest time and build up some (possibly rather complex) fallback layer that would try to "fix" problems with API response. We cannot really predict what could happen in the future. And (in my opinion) we should not care that much of making current Phragile versions compatible with older Phabricator releases - this is what we do Phragile releases for.
What should not happen is that Phragile crashes when API changes. By crashing I don't mean not showing changes in graph etc (what is what it did recently) but throwing exceptions all over the place, displaying some unpredictable errors etc. But this is not the case already, so I think we're fine.

> For non-Wikimedia users we could also add a hint in the readme, that Phabricator updates can be incompatible to the current Phragile version and should be tested before used.

It should be pretty clear that particular Phragile version only works with the specified Phabricator release. Not updating Phabricator will simply be "doing it on own risk" kind of thing. We should just be specific and clear in README and CHANGELOG.

jakobw:
I like the ideas and the approach in this PR a lot and I'm pretty sure having our own data structure for both transitions and tasks is the right way to go. I also like the ideas in the TODOs such as moving `createSnapshot` to its own class. 

Crazy idea: Since we're already storing everything in our own data structure we might not even need to store transitions anymore. Instead we could always save the latest status and the latest workboard column of each task as well as the time it was changed/moved. The advantage would be that we can probably clean up and throw away some code. On the other hand it would be additional work to implement and we might discover that some (future) feature ends up needing transaction.

About the problem with Phabricator updates:
It's a good idea to have something monitoring changes in the Phabricator API and I thought about taking the idea from @manicki one step further and instead of checking production snapshots we could run our acceptance tests against Phabricator master. That way we would get notified a couple of days before these changes would be deployed to the Wikimedia Phabricator instance. I chatted with Jan and Tobi about it and some ideas we had would be to either poke Travis, run it on a labs instance or on a Raspberry PI in the office which yells at us if something fails :)

manicki:
I'll give @jakobw's crazy idea (R) a try. If/Once I have it implemented I'll submit a competing pull request.

Running acceptance tests against Phabricator master sounds good. That should allow us to get warned really early as, as far as I know, Wikimedia Phabricator is kind of few weeks behind the master.

manicki:
I might be getting confused from looking into those Phabricator data too long but doesn't it sounds like a point against the idea of only storing most recent column/status data?
Let's assume there is a sprint with board with columns: "Backlog", "Doing", "Done", "Won't Do". In this project tasks moved to "Doing" or "Won't Do" columns are considered "done" (green in Phragile).
Let's assume there is a task "Fix all the bugs" in the sprint. Developer X grabs the task on May, 1st and moves it do "Doing". After spending some time on fixing several bugs she considers task finished and moves it to "Done" on May, 2nd. Next day Product Owner realizes there is a still an issue in the product and moves the task back to "Doing". Developer spends a week on investigating the issue and realizes it lies in the external component their product uses and they cannot fix it. After explaining the problem, she moves the task to "Done" (on May, 9th). Next morning the strict Product Owner who likes the order in her boards moves a task to "Won't Do" as the issue turned out to be outside of the scope of the product team (that happends May, 10th).
They way Phragile works currently, the task will be considered "done" on May, 9th. Moving it to the other "done" column the next day does not really affect taks's "closedness" for Phragile.
If we know only store the most recent column task was moved to, Phragile would correctly consider task "done" as it ended up in the "Won't Do" column but would consider wrong that it was closed on May, 10th.

Similar (maybe even more evident then above silly story) example when not using workboard columns to determine status of the task:
Let's say there is task X in the sprint. It gets looked into and it turns to be a duplicate of old good task A. So task X is closed (status changed to "resolved"). After some time more experience Phabricator user that loves the order marks the task X as a duplicate of task A. Phragile should still consider that task X was "closed"/"done" at the point it was "resolved". That would not be the case if only the most recent status changed was used by Phragile.

Those cases are probably not that crucial and for them to actually matter some "misusing" of Phabricator/Phragile must also happen but I am starting to think that maybe it still makes sense to have the whole task's column/status changes history stored in the data Phragile uses.

WMDE-Fisch:
/me votes for storing the transaction data. Seems to make sense looking at the examples given.

manicki:
Yeah, I'll be thinking a bit more about it if there are some more evident and obvious cases when we really need whole "transaction history". Those above are a bit of edge cases (although still possible; I am guess making Phragile only consider last "change to closed" was actually added based on some bug report).
After I am done with thinking I'd suggest to have task data in snapshots also stored in some Phragile's own structure (currently those are in a format that Phabricator API gives, with some not needed information) for consistency.
And then convert all the snapshots. Those two things do not seem too complex compared to the transaction data part.

jakobw:
Oh true, that's a valid example you brought up! I guess it could be worked around by having one field when a task was moved from an open status to a closed one and then another field for when it was moved to its current status but that sounds like a bad idea. So another +1 for storing transactions!

I think it's still good that we talked about the option. At least now we know why we didn't do it that way :D

manicki:
I have added a few commits (those might have been rather squashed into previous ones - please just let me know if your prefer me to do this) that are generally filling TODOs that have been left undone in the initial commit set.
When speaking about transaction data I'd consider this complete now.
Related things: storing only relevant task data in snapshots and an ultimate snapshot conversion script will be added in separate pull requests to make review easier (Github says it is already more than 1000 lines to review!)

app/Phragile/ColumnChangeTransaction.php
```diff
@@ -0,0 +1,72 @@
...
+	public function getNewColumnPHID()
+	{
+		return $this->newColumnPHID;
+	}
+
+	public function getTransactionData()
+	{
+		return [
+			'type' => 'columnChange', // TODO: this string should be moved to some constant,
+			//                        //so it there is no need to type it in in other places. But constant of what? Transaction::sth?
```
WMDE-Fisch: I would put it in each class. You could set it even in the base class and override it in the others. e.g.:
const PHAB_TYPE = 'undefined'; and then const PHAB_TYPE = 'columnChange';

app/Phragile/TransactionLoader.php
```diff
@@ -1,6 +1,7 @@
 <?php
 namespace Phragile;
 
+// TODO: rename to e.g. TransactionDataFetcher?
```
WMDE-Fisch: TransactionDataFetcher !

app/Phragile/TransactionFilter.php
```diff
@@ -1,6 +1,7 @@
 <?php
 namespace Phragile;
 
+// TODO: rename to e.g. TransactionDataFilter
```
WMDE-Fisch: TransactionDataFilter !

app/Phragile/TransactionRawDataProcessor.php
```diff
@@ -0,0 +1,95 @@
...
+				);
+			}
+		}
+		return $transactions;
+	}
+
+	private function isColumnChangeTransaction(array $rawData)
+	{
+		return array_key_exists('transactionType', $rawData)
+			&& $rawData['transactionType'] === 'core:columns';
```
WMDE-Fisch: These could also got to the Transaction class into a constant since you use them at least in two places. Maybe as PHAB_TYPE here.

app/Phragile/TransactionFilter.php
manicki: hmm, but shouldn't it actually be `TransactionRawDataFilter` as it deals with Phabricator's data (per `TransactionRawDataProcessor`)?

app/Phragile/TransactionLoader.php
manicki: hmm, but shouldn't it actually be `TransactionRawDataFetcher` as it deals with Phabricator's data (per `TransactionRawDataProcessor`)?

app/Phragile/TransactionRawDataProcessor.php
manicki: I don't think this belongs to Transaction class, it is Phabricator-specific.
I think this is only line of production code that mentions `core:columns` (some test data might be also doing it which is expected I think). And if not only then only Phabricator2Phragile transaction data conversion code should know anything on how currently Phabricator represents column change transactions (and any other kind of transactions).
Any way, be it he only line explicitly mentioning `core:columns` it might make sense to make those Phabricator transaction type string class constants so they are changed in one place no matter how the processor's code evolves. At least I think it might help a bit, so I'll change it. Of course it would only be valuable unless next change of transaction data structure in Phab API will be making it all look completely differently but this is rather unlikely.

app/Phragile/ColumnChangeTransaction.php
manicki: yes, this would be better than adding some consts to the abstract Transition class, a const for each extending class. A class shouldn't really know about its "children" I think

app/Phragile/TransactionRawDataProcessor.php
manicki: oh, of course the phrabricator transaction type string is also used explicitly in `TransactionFilter`, that's what you meant, sorry! Forget what I was mumbling about in the above comment.
Then moving the string to a constant makes sense. But I'd definitely keep it away from `*Transaction` classes

manicki:
I am closing this pull request and re-submit it against a "new snapshot structure" branch. This is done in order to have all the code affecting existing snapshots in one place so the state of master with only partly changed snapshots (e.g. only transaction part changed) will not get deployed untill ALL is ready.

app/Phragile/TransactionFilter.php
WMDE-Fisch: +1 TransactionRawDataFilter

app/Phragile/TransactionLoader.php
WMDE-Fisch: +1 TransactionRawDataFetcher

app/Phragile/Transaction.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Phragile;
+
+abstract class Transaction {
+
+	const TYPE = 'undefined';
+
+	// TODO: consider renaming the field to something more meaningful
```
WMDE-Fisch: IMO timestamp is good.

app/Phragile/Transaction.php
```diff
@@ -0,0 +1,20 @@
...
+	protected $timestamp;
+
+	public function getTimestamp()
+	{
+		return $this->timestamp;
+	}
+
+	// TODO: would getSnapshotData be a better name?
```
WMDE-Fisch: why not just getData ?

PR #241, opened 2016-05-11
Use simplified task data structure

manicki:
Hopefully no too long pull request description this time (see instead a detailed commit message of https://github.com/wmde/phragile/commit/1e8eacda65627210e953da1157d401de08727f08 describing what is the general idea of the change).

Basically this is aiming to do similar thing that https://github.com/wmde/phragile/pull/240 does for transactions. There has already been some abstraction for tasks but it has been mixing both "general" task data and UI-specifics which I am trying to separate here.

This depends on https://github.com/wmde/phragile/pull/240 and contains commits from it. Commits relevant for this PR start from https://github.com/wmde/phragile/commit/1e8eacda65627210e953da1157d401de08727f08

As mentioned in https://github.com/wmde/phragile/commit/1e8eacda65627210e953da1157d401de08727f08 there will be a follow up with snapshot data conversion script based on this PR and https://github.com/wmde/phragile/pull/240

As could be easily noticed I had some serious naming problems here. I'll very much appreciate any better name suggestions! Especially I am not happy with calling things "rendering" here (render does not exactly mean "prepare for presentation", does it?). I couldn't come up with any other short and meaningful name. Please help!


manicki:
Ugh, Travis is lagging?

manicki:
Force-pushed after adding a commit to https://github.com/wmde/phragile/pull/240

manicki:
I am closing this pull request and re-submit it against a "new snapshot structure" branch. This is done in order to have all the code affecting existing snapshots in one place so the state of master with only partly changed snapshots (e.g. only transaction part changed) will not get deployed untill ALL is ready.

PR #243, opened 2016-05-11
Use simplified transaction data structure internally and in snapshots

manicki:
This is a re-submit of https://github.com/wmde/phragile/pull/240 but onto "feature branch"
Please refer to the original PR for discussion on in-line comments on code.


manicki:
@WMDE-Fisch: renamed `Transaction::getTransactionData` to `getData` as you suggested in https://github.com/wmde/phragile/pull/240

app/Phragile/MergedIntoTransaction.php
```diff
@@ -0,0 +1,22 @@
...
+class MergedIntoTransaction extends Transaction {
+
+	const TYPE = 'mergedInto';
+
+	public function __construct($timestamp)
+	{
+		$this->timestamp = $timestamp;
+	}
+
+	public function getData()
```
jakobw: So I just stared at this class for 5 minutes (exaggerated) and had no idea what it's for and why we'd want to store MergedIntoTransactions without storing what got merged into what. I then dug into the code a bit deeper and remembered that merging a task is basically the same for Phragile as closing it.
Maybe we can add a little explanation either above the method or above the class to avoid future confusion . What do you think?

manicki: You're absolutely right. I've also first forgotten about it, and only realized that when replacing old code and spotted this "mergedinto". I'll add an explanation. Plus I think it would even make sense to rename this class to something like `MergeAndCloseTransaction` or `CloseByMergingTransaction`. Both names are not very appealing but maybe I'll manage to come up with something better while going through a rebase hell :)

jakobw:
Aside from the small comment and little bits of commented out code (not sure if we're at the nitpick level of review yet) I'm fine with merging this into the giant feature branch :D

PR #244, opened 2016-05-11
Use simplified task data structure

manicki:
This is a re-submit of https://github.com/wmde/phragile/pull/241 but onto "feature branch"
Please refer to the original PR for discussion on in-line comments on code.

Please note that this contains a commit from https://github.com/wmde/phragile/pull/248.
The actual changes starts at https://github.com/wmde/phragile/pull/244/commits/467d755b57f6858d4935bc5d55ef2439b5f6e58e


manicki:
Rebased after adding a commit to https://github.com/wmde/phragile/pull/243 and renamed `Task::getTaskData` to `getData`.

app/Phragile/RenderedTask.php
```diff
@@ -0,0 +1,119 @@
+<?php
+namespace Phragile;
+
+class RenderedTask {
```
jakobw: Unhelpful comment: Looking at the `Task` class I found it a bit annoying how similar these two are. I see that they each have their purpose but something does not seem right about this.

I will finish scanning the entire PR and try to come up with a more constructive comment/suggestion.

app/Phragile/StatusByStatusFieldDispatcher.php
```diff
@@ -9,17 +9,18 @@ public function __construct($reviewTagPHID)
...
 
-	private function isTaskInReview(array $task)
+	private function isTaskInReview(Task $task)
 	{
-		return !$this->isClosed($task) && in_array($this->reviewTagPHID, $task['attachments']['projects']['projectPHIDs']);
+		return !$this->isClosed($task) && in_array($this->reviewTagPHID, $task->getProjectPHIDs());
 	}
 
-	private function isTaskBeingDone(array $task)
+	// TODO: isDone($task)?
```
jakobw: I realize now how terrible the current name is. It actually mean that the task is in doing. So `isDone` is wrong but `isInDoing` might be better than the current version?

app/Phragile/StatusByWorkboardDispatcher.php
```diff
@@ -2,6 +2,7 @@
 namespace Phragile;
 
 class StatusByWorkboardDispatcher implements StatusDispatcher {
+	// TODO: $transactions field not needed?
```
jakobw: Remooooove!

app/Phragile/StatusByStatusFieldDispatcher.php
manicki: Right, my mistake. Removed TODO

app/Phragile/StatusByWorkboardDispatcher.php
manicki: done

app/Phragile/RenderedTask.php
manicki: True. So what I did now is introducing Domain and Presentation namespaces and having two Task classes in each namespace.

PR #245, opened 2016-05-11
Ultimate snapshot data migration command

manicki:
My idea was to change the existing `migrate:snapshots` command so it converts transaction data and task data from both `maniphest.query` and `maniphest.search` response formats.

I am assuming snapshot structure changes are going to be released in version 3.0.0 and added migration instruction to `UPGRADE.md`. Some description should probably be also added to `README` and `RELEASE-NOTES` but I'd rather do it separately.

### DIFF
--- a/app/Console/Commands/lib/SnapshotTaskDataConverter.php
+++ b/app/Console/Commands/lib/SnapshotTaskDataConverter.php
@@ -2,16 +2,40 @@
 
 namespace App\Console\Commands\Lib;
 
+use Phragile\Domain\Task;
+use Phragile\TaskRawDataProcessor;
+
 class SnapshotTaskDataConverter {
 
 	public function convert(array $taskData)
 	{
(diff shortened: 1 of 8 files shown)


app/Console/Commands/lib/SnapshotTaskDataConverter.php
```diff
@@ -0,0 +1,93 @@
...
+		return $fields;
+	}
+
+	/**
+	 * Format of old custom field was e.g. std:maniphest:WMDE:story_points or isdc:sprint:storypoints.
+	 * Equivalent new format would be custom.WMDE_story_points or custom.storypoints.
+	 * @param string $s
+	 * @return string
+	 */
+	private function strAfterSecondColon($s)
```
jakobw: Not super important but the name doesn't really fit anymore now. Could be something like `extractFieldName` maybe?

manicki: good point. will change

jakobw:
Looks really good!

PR #247, opened 2016-05-12
Prepare Phragile 3.0.0 release

manicki:
I am submitting this already so that Phabricator "version number" that changed the column change transaction structure does not get forgotten (it always takes me a while to browse through those Phabricator release notes).

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -27,7 +27,7 @@ To support newer versions of Phabricator, Phragile uses manifest.search instead 
 
 ## From 1.0 release to 1.1
 
-Phragile 1.1. starts using Conduit API Tokens instead of the deprecated Conduit API Certificate. 
+Phragile 1.1. starts using Conduit API Tokens instead of the deprecated Conduit API Certificate.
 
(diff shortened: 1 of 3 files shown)


jakobw:
Taking over this PR by @manicki for https://phabricator.wikimedia.org/T135712

UPGRADE.md
```diff
@@ -7,6 +7,14 @@ When upgrading Phragile to a newer version **always** run
 
 after updating the files in the root directory of your application.
 
+## From 2.0.0 release to 3.0.0
```
manicki: This is probably fine although when previously using version 1.0 or 1.1, and then getting 3.0.0 (ie. skipping 2.0.0) `snapshots:migrate` will also convert snapshots! So there is no intermediate step needed (first update to 2.0.0, and then to 3.0.0). I am not sure if this is intended to allow such a "super upgrade", though.

jakobw: Stupid me, I just saw that you already wrote a note for this case in the UPGRADE.md and it just wasn't in this branch yet :)
Updated!

